PR #32, opened 2018-03-05
Convert C# -> VB trivia (comment and newline)

### LINKED ISSUE #8: C# -> VB code comments are not converted
It looks like they are just ignored.

GrahamTheCoder:
Fixes #8 (or rather will fix it once this PR is complete)

Simple cases will be handled to some extent, but need to fix:
* Spacing issues now that whitespace/newline trivia is ported
* Structures of different shapes, moving their trivia to the right place
  * e.g. A class definition
  * e.g. Trivia gets attached to the thing at the end of the statement, which in C# is a semicolon, but in VB will be something else
* Block comments mid-line in C# that have no equivalent in VB

I don't have time to work on this right now, but I'm leaving this here for others to continue the work if so desired. Understanding the test failures should be a good start, though the test setup in CodeConverterBase will need some changes to test for the correct behaviour.

### DIFF
--- a/CodeConverter.sln
+++ b/CodeConverter.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2010
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29111.141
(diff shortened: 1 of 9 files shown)


GrahamTheCoder:
As mentioned in #506 I'm planning an overhaul of comment conversion. It should be far easier to support C# to VB with the planned system since it will require fewer special cases

PR #43, opened 2018-03-13
VbToCs: integer divide, narrowing/widening

BaronBodissey:
I don't really know what I'm doing...

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -565,14 +565,31 @@ namespace ICSharpCode.CodeConverter.CSharp
                 var block = node.OperatorStatement;
                 var attributes = block.AttributeLists.SelectMany(ConvertAttribute);
                 var modifiers = ConvertModifiers(block.Modifiers, TokenContext.Member);
-                return SyntaxFactory.OperatorDeclaration(
-                    SyntaxFactory.List(attributes),
-                    modifiers,
-                    (TypeSyntax)block.AsClause?.Type.Accept(TriviaConvertingVisitor) ?? SyntaxFactory.PredefinedType(SyntaxFactory.Token(SyntaxKind.VoidKeyword)), ConvertToken(block.OperatorToken),
-                    (ParameterListSyntax)block.ParameterList.Accept(TriviaConvertingVisitor),
-                    SyntaxFactory.Block(node.Statements.SelectMany(s => s.Accept(CreateMethodBodyVisitor()))),
-                    null
-                );
+                SyntaxToken? conversionOp = null;
+                conversionOp = modifiers.FirstOrNullable(t=>VisualBasicConverter.IsConversionOperator(t));
+                if (conversionOp.HasValue) {
+                    modifiers=modifiers.Remove(conversionOp.Value);
+                    return SyntaxFactory.ConversionOperatorDeclaration(
+                       SyntaxFactory.List(attributes),//attributes
+                       modifiers,//modifiers
+                       conversionOp.Value,//implicitOrExplicit
+                       (TypeSyntax)block.AsClause?.Type.Accept(TriviaConvertingVisitor) ?? SyntaxFactory.PredefinedType(SyntaxFactory.Token(SyntaxKind.VoidKeyword)),//type
+                       (ParameterListSyntax)block.ParameterList.Accept(TriviaConvertingVisitor),//parameterList
+                       SyntaxFactory.Block(node.Statements.SelectMany(s => s.Accept(CreateMethodBodyVisitor()))),//body
+                       null//expressionBody
+                   );
+                   
+                } else {
+                    return SyntaxFactory.OperatorDeclaration(
+                        SyntaxFactory.List(attributes),//attributes
+                        modifiers,//modifiers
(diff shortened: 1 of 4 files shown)


ICSharpCode.CodeConverter/CSharp/VisualBasicConverter.cs
```diff
@@ -333,6 +341,7 @@ static SyntaxKind ConvertToken(VBasic.SyntaxKind t, TokenContext context = Token
                 case VBasic.SyntaxKind.MultiplyExpression:
                     return SyntaxKind.MultiplyExpression;
                 case VBasic.SyntaxKind.DivideExpression:
+                case VBasic.SyntaxKind.IntegerDivideExpression:
```
GrahamTheCoder: Don't worry about integer divide - that's already fixed in master

Tests/CSharp/ExpressionTests.cs
```diff
@@ -633,5 +633,62 @@ public string TestMethod()
...
+    {
+        return new MyInt();
+    }
+    public static implicit operator int
+(MyInt myInt)
+    {
+        return 1;
+    }
+}");
+        }
```
GrahamTheCoder: Yep you're on the right track!

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -565,14 +565,31 @@ public override CSharpSyntaxNode VisitOperatorBlock(VBSyntax.OperatorBlockSyntax
...
+                } else {
+                    return SyntaxFactory.OperatorDeclaration(
+                        SyntaxFactory.List(attributes),//attributes
+                        modifiers,//modifiers
+                        (TypeSyntax)block.AsClause?.Type.Accept(TriviaConvertingVisitor) ?? SyntaxFactory.PredefinedType(SyntaxFactory.Token(SyntaxKind.VoidKeyword)),//returnType
+                        ConvertToken(block.OperatorToken),//operatorToken
+                        (ParameterListSyntax)block.ParameterList.Accept(TriviaConvertingVisitor),//parameterList
+                        SyntaxFactory.Block(node.Statements.SelectMany(s => s.Accept(CreateMethodBodyVisitor()))),//body
+                        null//expressionBody
+                    );
```
GrahamTheCoder: Looks like you're on the right track here

GrahamTheCoder:
I've just had a quick look at the diff - in the next few days I'll check it out locally and see what needs doing to get this mergeable.

Tests/CSharp/ExpressionTests.cs
```diff
@@ -633,5 +633,62 @@ public string TestMethod()
...
+End Class"
+, @"class TestClass
+{
+    public void TestMethod(ref int i)
+    {
+        int k;
+        k = i / 1000000;
+    }
+}");
+        }
```
GrahamTheCoder: Yep this is the right sort of test for integer division - now already covered in master by https://github.com/icsharpcode/CodeConverter/blob/4f6bcb3ff7213e321dd73bc3e9e507956e44244e/Tests/CSharp/ExpressionTests.cs#L84

Tests/CSharp/ExpressionTests.cs
BaronBodissey: Here there's an issue I didn't knew how to fix: the parameters list is send to the next line instead of staying on the same line as the operator declaration.
It seems it's roslyn which generate this line feed at
*resultNode.ToFullString()*
# projectconvertion.cs line 61
I tried to find any option to prevent this to happen but didn't find any. Maybe you already know the answer. Indeed I had to use the TestConversionVisualBasicToCSharpWithoutComments instead of the TestConversionVisualBasicToCSharp which is clearly failing because of this evil line feed :-).

BaronBodissey:
@GrahamTheCoder Also, as you requested I created a specific issue for this contribution
#43 .
Can you confirm that it is what was needed.
AlsoAlso, I'm not sure what is the process in here, so I don't know how can I update my fork with the main repo. If you could explain  it to me thanx. :-)

Tests/CSharp/ExpressionTests.cs
GrahamTheCoder: Yes I'm happy with that fix for the test failure. Good job figuring it out - I've added this as a separate issue to try to make it clearer for others (#45).

GrahamTheCoder:
Github's docs give the git commands - in this case, the icsharpcode repo is referred to as the "upstream" which may help in finding other related answers:
https://help.github.com/articles/merging-an-upstream-repository-into-your-fork/

Hope that helps

Tests/CSharp/ExpressionTests.cs
GrahamTheCoder: I think this test probably belongs in `MemberTests` since it's really about a special kind of class member. Would you mind moving it into there?

BaronBodissey: Will do...if you can wait...I'm on the steep slope of the learning curve ;-)

GrahamTheCoder: @BaronBodissey No worries - I'll tidy things up and get it merged now

GrahamTheCoder:
Great first pull request, thanks a lot!

PR #116, opened 2018-04-26
Netstandard 2.0

### LINKED ISSUE #89: VB -> C#: Increase number of default global imports for web conversion
* [ ] Look at a few VB projects around the web and find any common (BCL) global imports to add

Essentially the converter will do better the more (non-conflicting) things are imported, so picking all the standard ones seems sensible. For example `System.Runtime.CompilerServices` https://github.com/icsharpcode/CodeConverter/issues/88#issuecomment-379787915

### LINKED ISSUE #96: DefaultMetadataReferences are different in CodeWithOptions and DiagnosticTestBase
Shouldn't the Default Metadata References be the same in both places?  If so then CodeWithOptions can be updated and DiagnosticTestBase can be removed.

GrahamTheCoder:
Fixes #89
Fixes #96 

* [ ] Figure out how to get AppVeyor to restore nuget packages correctly and build net standard 2.0


GrahamTheCoder:
Thanks I'll take a look

PR #137, opened 2018-06-26
Closes #136 - VB -> C# Use parentheses around single parameter in lambda when parameter has explicit type

### LINKED ISSUE #136: VB -> C#: Lambda expression missing parentheses for single parameter with explicit type
### Input code
```vbnet
Dim f As Func(Of Integer, Integer) = Function(x As Integer) x
```

### Erroneous output
```C#
Func<int, int> f = int x => x;
```

### Expected output
```C#
Func<int, int> f = (int x) => x;
```

CoderNate:
There was an existing test case that showed the issue so this pull request updates that test and doesn't add any new tests.


GrahamTheCoder:
Great stuff - as you can see there are a fair few issues so your efforts really are appreciated!

PR #140, opened 2018-06-26
Closes #139 - VB -> C#: Array literals not always converted to implicit C# array

### LINKED ISSUE #139: VB -> C#: Array literals not always converted to implicit C# array
Currently a VB array literal is only converted to a C# implicit array if the Roslyn semantic model knows that the expression type is an array.  Otherwise the array literal is left as a stray collection initializer.

### Input code
```vbnet
    Private Sub DoStuff(a As Object)
    End Sub
    Private Sub TestMethod()
        DoStuff({1, 2})
    End Sub
```

### Erroneous output
```C#
DoStuff({ 1, 2 });
```

### Expected output
```C#
DoStuff(new[] { 1, 2 });
```

CoderNate:
I checked all occurrences of 'CollectionInitializer' in the [VB.NET grammar](https://docs.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/expressions) to try and make sure that there isn't any other node that a CollectionInitializer could belong to where you _wouldn't_ turn it into an implicit C# array.

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -1162,11 +1162,13 @@ namespace ICSharpCode.CodeConverter.CSharp
 
             public override CSharpSyntaxNode VisitCollectionInitializer(VBSyntax.CollectionInitializerSyntax node)
             {
-                var typeInfo = _semanticModel.GetTypeInfo(node);
-                var isArrayType = typeInfo.ConvertedType?.IsKind(SymbolKind.ArrayType) == true;
-                var initializerType = isArrayType ? SyntaxKind.ArrayInitializerExpression : SyntaxKind.CollectionInitializerExpression;
+                var isArrayLiteral =
+                    !( node.Parent is VBSyntax.ObjectCollectionInitializerSyntax
+                    || node.Parent is VBSyntax.CollectionInitializerSyntax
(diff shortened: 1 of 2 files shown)


ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -1162,11 +1162,13 @@ public override CSharpSyntaxNode VisitArrayCreationExpression(VBSyntax.ArrayCrea
 
             public override CSharpSyntaxNode VisitCollectionInitializer(VBSyntax.CollectionInitializerSyntax node)
             {
-                var typeInfo = _semanticModel.GetTypeInfo(node);
-                var isArrayType = typeInfo.ConvertedType?.IsKind(SymbolKind.ArrayType) == true;
-                var initializerType = isArrayType ? SyntaxKind.ArrayInitializerExpression : SyntaxKind.CollectionInitializerExpression;
+                var isArrayLiteral =
+                    !( node.Parent is VBSyntax.ObjectCollectionInitializerSyntax
+                    || node.Parent is VBSyntax.CollectionInitializerSyntax
+                    || node.Parent is VBSyntax.ArrayCreationExpressionSyntax);
```
GrahamTheCoder: Looks good! I just inverted this logic to simplify it slightly in https://github.com/icsharpcode/CodeConverter/commit/38dda0bf98d22e24181e2c2df8e7f51b13b24f3b

PR #148, opened 2018-06-27
Closes #147 VB -> C#: Conversion error for indexer on property value of unresolved type

### LINKED ISSUE #147: VB -> C#: Conversion error for indexer on property value of unresolved type
Get a "Cannot Convert" error comment in converted code when using an indexer on the value returned from a property with an unresolved type.  By unresolved I mean Roslyn fails to resolve the type due to a missing DLL or project reference, so no information is available about it in the semantic model.

### Input code
```vbnet
Class TestClass
    Public Property SomeProperty As System.Some.UnknownType
    Private Sub TestMethod()
        Dim value = SomeProperty(0)
    End Sub
End Class
```

### Erroneous output
```C#
/* Cannot convert:  System.ArgumentOutOfRangeException ... 
```

### Expected output
```C#
class TestClass
{
    public System.Some.UnknownType SomeProperty { get; set; }
    private void TestMethod()
    {
        var value = SomeProperty(0);
    }
}
```

CoderNate:
In VB, properties can have indexers like this:
```vbnet
Public Property SomeProp(index As Integer) As Integer
```
When running the provided test case, VisitSimpleArgument gets an ArgumentOutOfRangeException on the call to:
```C#
parameterKinds[argId]
```
Here's why.  VisitSimpleArgument calls GetInvocationSymbol on the invocation involving SomeProperty which returns a symbol with information about that property.  The current program looks like it was mostly written with method invocation in mind so it's trying to determine if the method parameter corresponding to the argument being passed is a ref or out parameter.  In the test case, the invocation is trying to target the value _returned by_ the property (like if the value was an array or a list) as opposed to calling an indexer on the property itself.  But instead, GetParameters is called for the "SomeProperty" symbol which doesn't have an indexer and therefore has zero parameters.  So you get the ArgumentOutOfRangeException.

When I tried declaring SomeProperty as an array type instead of a type that Roslyn couldn't resolve, I got a null value back from GetInvocationSymbol.  I'm not sure why that is.

### DIFF
--- a/Tests/CSharp/ExpressionTests.cs
+++ b/Tests/CSharp/ExpressionTests.cs
@@ -377,7 +377,7 @@ End Class", @"class TestClass
     public System.Some.UnknownType SomeProperty { get; set; }
     private void TestMethod()
     {
-        var value = SomeProperty(0);
+        var value = SomeProperty[0];
     }
 }");
         }
--- a/Tests/CSharp/ExpressionTests.cs
+++ b/Tests/CSharp/ExpressionTests.cs
@@ -365,6 +365,24 @@ class TestClass
         }
 
         [Fact]
+        public void InvokeIndexerOnPropertyValue()
+        {
+            TestConversionVisualBasicToCSharp(@"Class TestClass
+    Public Property SomeProperty As System.Some.UnknownType
+    Private Sub TestMethod()
+        Dim value = SomeProperty(0)
+    End Sub
+End Class", @"class TestClass
+{
+    public System.Some.UnknownType SomeProperty { get; set; }
+    private void TestMethod()
+    {
+        var value = SomeProperty(0);
+    }
+}");
+        }
+
+        [Fact]
         public void ExternalReferenceToOutParameter()
         {
             TestConversionVisualBasicToCSharp(@"Class TestClass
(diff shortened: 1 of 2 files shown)


Tests/CSharp/ExpressionTests.cs
```diff
@@ -364,6 +364,24 @@ private void TestMethod(string str)
...
+    Public Property SomeProperty As System.Some.UnknownType
+    Private Sub TestMethod()
+        Dim value = SomeProperty(0)
+    End Sub
+End Class", @"class TestClass
+{
+    public System.Some.UnknownType SomeProperty { get; set; }
+    private void TestMethod()
+    {
+        var value = SomeProperty(0);
```
GrahamTheCoder: I was expecting this to say `SomeProperty[0]` given the test name.

GrahamTheCoder:
The code currently makes reasonably little effort to handle missing semantic information and I think this will be a great addition since it's such a common case the code gets wrong at the moment in that case. 

In the general case where type information is missing it's impossible to be *sure* whether parentheses or square brackets are needed because the unknown object could be a delegate. However the chances of having an unknown delegate stored as a field/local seem lower than having an unknown non-delegate type with an indexer stored , so for a standalone identifier I think we should err on the side of assuming it's an indexer - which I think is what you're trying to do here.

In general, I'm trying to only write tests that have full semantic information. I'm slowly correcting errors in the old ones by flipping the condition on ProjectConversion.cs:95 and following the test errors. At some point I'd like to have the test harness always report syntax errors, so I think it'd make sense to have a separate folder for tests which intentionally deal with a broken semantic model. We should probably add a bit of rationale I did above when doing so, since it's much harder to understand what tradeoffs are being made in the face of ambiguous input.

Issue #107, opened 2018-04-19
VB -> C#: Differing namespace rules in C#


GrahamTheCoder:
A user review said "C# files require namespaces, but the tool did not add them" (see Tony Waterman's [review](https://marketplace.visualstudio.com/items?itemName=SharpDevelopTeam.CodeConverter#review-details))

In general a C# doesn't need to be explicitly in a namespace, but I think there likely are some differences between namespaces to be investigated and fixed. Any source code that ends up not compiling due to needing a namespace would be much appreciated.

[Type promotion](https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/declared-elements/type-promotion) in VB may cause some unexpected errors in conversion

CoderNate:
There are a few differences in VB.
1)  Type promotion, like you said.  You could use the C# 6 "using static" for that.
2)  The "Root Namespace" defined in the project's properties gets automatically inserted at the beginning of every top-level declaration in the source code.  So if the top level item for a file is a namespace declaration, the project's root namespace will be added to the beginning.  If the top level is a module or a class, the root namespace will be added to that.  In order to to declare something outside of the root namespace you can start it with "Global" like Global.Some.Namespace.
3)  VB lets you declare partially-qualified types based on namespaces you've imported:
```vbnet
Imports System.Collections
...
Dim d As Specialized.OrderedDictionary
```
You would have to do this in C#:
```C#
using Specialized = System.Collections.Specialized;
...
Specialized.OrdereredDictionary d;
```

GrahamTheCoder:
1. `QualifyNode` attempts to deal with point 2, by fully qualifying partially qualified namespaces - though it's a bit clunky and there will be some bugs.
2. Is probably not dealt with by the above. It could be extended to do so, or we could use the `using static` approach. I'm happy with either.
3. Your PR #150 deals with it entirely 👍

GrahamTheCoder:
@CoderNate Looking at `QualifyNode` now, can you see any other cases that might not be handled or deserve a test case before we can close this issue?

CoderNate:
I just created pull request #156 that fixes a bug I encountered.  I mentioned in the comment on that pull request that it might be good to have a test for nested classes.  But otherwise I can't think of anything else.  Nice work!

GrahamTheCoder:
I believe this is now fixed 🎉

PR #150, opened 2018-06-27
#107 VB -> C#: Add VB root namespace to all converted files

CoderNate:
Warning, this will break almost all test cases by enclosing generated code in a namespace.  All the unit tests use an instance of VisualBasicCompilationOptions that uses .WithRootNamespace("TestProject").  Not sure if you want to just change it to use a blank root namespace or how you'd want to handle that.

Anyway, this pull request changes all of a projects generated files as follows:
- if the only member declaration is a namespace declaration, it prepends it with the VB root namespace
- otherwise all member declarations are wrapped in a new top-level namespace

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -142,7 +142,7 @@ namespace ICSharpCode.CodeConverter.CSharp
                     return SyntaxFactory.QualifiedName(PrependName(qName.Left, toPrepend), qName.Right);
                 }
                 else {
-                    throw new Exception("Unknown name syntax");
+                    throw new ArgumentOutOfRangeException(nameof(name), name, $"{name.GetType().Name} of kind {name.Kind()} not expected within namespace declaration");
                 }
             }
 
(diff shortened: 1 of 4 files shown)


GrahamTheCoder:
Whoops didn't add a test for this. I've seen it in action and it looks good to me though.
If you fancy adding a test, you can use the `SolutionAndProjectTests` style of comparing to files on disk.

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -105,7 +105,7 @@ public override CSharpSyntaxNode VisitCompilationUnit(VBSyntax.CompilationUnitSy
                 }
 
                 var attributes = SyntaxFactory.List(node.Attributes.SelectMany(a => a.AttributeLists).SelectMany(ConvertAttribute));
-                var convertedMembers = node.Members.Select(m => (MemberDeclarationSyntax)m.Accept(TriviaConvertingVisitor));
+                var convertedMembers = node.Members.Select(m => (MemberDeclarationSyntax)m.Accept(TriviaConvertingVisitor)).ToReadOnlyCollection();
```
CoderNate: :thumbsup:  Yeah I knew it was wrong when I did the goofy looking Skip(1).Any() in order to avoid doing a Count() that enumerates the whole list, but for some reason I didn't fix it...

GrahamTheCoder: IReadOnlyCollection is my best friend

Issue #152, opened 2018-06-29
VB -> C#: Method invocation of inherited method gets qualified with base class


CoderNate:
### Input code
```vbnet
Class TestClassBase
    Overridable Sub DoStuff()
    End Sub
End Class
Class TestClass
    Inherits TestClassBase
    Private Sub TestMethod()
        DoStuff()
    End Sub
End Class
```

### Erroneous output
```C#
class TestClassBase {
    public virtual void DoStuff()
    {}
}
class TestClass : TestClassBase {
    private void TestMethod() {
        TestClassBase.DoStuff();
    }
}
```

### Expected output
```C#
class TestClassBase {
    public virtual void DoStuff()
    {}
}
class TestClass : TestClassBase {
    private void TestMethod() {
        DoStuff();
    }
}
```

In the NodeVisitor class the QualifyNode method is responsible for this.  If 'DoStuff' is declared in TestClass instead of TestClassBase, the following code would strip off the 'TestClass' part from the qualified name, but it's not checking for inherited class names:
```C#
    var firstMatch = prefixes.FirstOrDefault(p => qualifiedName.StartsWith(p));
    if (firstMatch != null)
    {
        // CSharp allows partial qualification within the current type's parent namespace
        qualifiedName = qualifiedName.Substring(firstMatch.Length);
    }
```

PR #153, opened 2018-06-29
#152 VB -> C#: Fix QualifyNode to handle inheritance

CoderNate:
I think this fixes the problem, but I don't know my way around the Roslyn type symbols too well.
EDIT:  This breaks the PartiallyQualifiedName test because the while loop I added keeps checking base classes all the way to System.Object in which case the enumerable of prefixes will contain "System".


GrahamTheCoder:
This seems like a good approach, sorry I haven't had time to properly get into reviewing it yet. When I do I'll figure out something to do about the broken test.

GrahamTheCoder:
Sorry this took such a long time, but now we're rid of my original clunky code that I wrote when *I* didn't understand Roslyn type symbols too well, so hopefully future issues in this area will be more easily fixed.

PR #155, opened 2018-07-15
vb -> C# Add parentheses around conditional expression in string interpolation

### LINKED ISSUE #154: vb -> C# Compiler error with Conditional Operator in String Interpolation
### Input code
```vbnet
Public Function GetString(yourBoolean as Boolean) As string
   Return $"You {if(yourBoolean, "do", "do not")} have a true value"
End Function
```

### Erroneous output
```c#
public string GetString(bool yourBoolean)
{
   return $"You {yourBoolean ? "do" : "do not"} have a true value";
}
```

### Expected output
```c#
public void GetString(bool yourBoolean)
{
   return $"You {(yourBoolean ? "do" : "do not")} have a true value";
}
```

### Details
Both extension and web app. 

I receive the following error in this situation

> Error	CS8361	A conditional expression cannot be used directly in a string interpolation because the ':' ends the interpolation. Parenthesize the conditional expression.	Tests
 

Version in use: 5.8.0.0

I am going to try to contribute a fix on this, but if I fail on it, I will ask for help

dan-drews:
Fixes #154 

### Problem
There was a compiler error when using a conditional operator in String Interpolation

### Solution
* I took a look at the other code in NodeVisitor.cs, and it appears that it is common to conditionally change the return information based upon certain situations
* I am most concerned about 2 parts of this:
  * `node.Parent.GetType() == typeof(VBSyntax.InterpolationSyntax)` - Is there a better way to do this?
  * Location of my test. Does it belong in a better place?

* [x] All tests pass - The tests did not pass, but when I reverted my changes, the same tests failed, so I do not believe the tests were previously passing, so I will check this off. If this is wrong, please let me know

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -1376,11 +1376,16 @@ namespace ICSharpCode.CodeConverter.CSharp
 
             public override CSharpSyntaxNode VisitTernaryConditionalExpression(VBSyntax.TernaryConditionalExpressionSyntax node)
             {
-                return SyntaxFactory.ConditionalExpression(
+                var expr = SyntaxFactory.ConditionalExpression(
                     (ExpressionSyntax)node.Condition.Accept(TriviaConvertingVisitor),
                     (ExpressionSyntax)node.WhenTrue.Accept(TriviaConvertingVisitor),
                     (ExpressionSyntax)node.WhenFalse.Accept(TriviaConvertingVisitor)
                 );
+
+                if (node.Parent.GetType() == typeof(VBSyntax.InterpolationSyntax))
+                    return SyntaxFactory.ParenthesizedExpression(expr);
+                
+                return expr;
             }
 
             public override CSharpSyntaxNode VisitTypeOfExpression(VBSyntax.TypeOfExpressionSyntax node)
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -1382,7 +1382,7 @@ namespace ICSharpCode.CodeConverter.CSharp
                     (ExpressionSyntax)node.WhenFalse.Accept(TriviaConvertingVisitor)
                 );
 
-                if (node.Parent.GetType() == typeof(VBSyntax.InterpolationSyntax))
+                if (node.Parent is VBSyntax.InterpolationSyntax)
                     return SyntaxFactory.ParenthesizedExpression(expr);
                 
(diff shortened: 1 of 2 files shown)


ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -1376,11 +1376,16 @@ public override CSharpSyntaxNode VisitBinaryConditionalExpression(VBSyntax.Binar
...
             {
-                return SyntaxFactory.ConditionalExpression(
+                var expr = SyntaxFactory.ConditionalExpression(
                     (ExpressionSyntax)node.Condition.Accept(TriviaConvertingVisitor),
                     (ExpressionSyntax)node.WhenTrue.Accept(TriviaConvertingVisitor),
                     (ExpressionSyntax)node.WhenFalse.Accept(TriviaConvertingVisitor)
                 );
+
+                if (node.Parent.GetType() == typeof(VBSyntax.InterpolationSyntax))
+                    return SyntaxFactory.ParenthesizedExpression(expr);
```
GrahamTheCoder: To do this sort of check it's generally preferential to use `node.Parent is VBSyntax.InterpolationSyntax`. This is on the assumption that if a future version of the library had a subclass, that we'd likely also want the same behaviour for that subclass.

In the case of this specific library, there's a special method for these sorts of checks because they're so commonly needed, which provides an abstraction layer over the type system. So you can call something like `node.Parent.IsKind(SyntaxKind.InterpolatedStringExpression)`.

GrahamTheCoder:
Incidentally, I imagine this same sort of bug will exist elsewhere in the code (where we change precedence), but the code is in a separate place at the moment. I could imagine in future passing every expression conversion through some kind of single precedence correction point, but I think that should be out of scope until the next time this kind of bug bites.

I also realize that interpolation is a bit special, and it may not be precedence that's directly causing this problem you observed, but since the same line of code could correct both problems, it seems sensible to do so now.

dan-drews:
Thanks so much for the feedback on this @GrahamTheCoder. This is really my first attempt at doing anything that is related to the Microsoft.CodeAnalysis namespace, so to be completely honest, I am not fully understanding all of the things you are referring to, this bug happened to cause an issue on a file I was converting, so I did my best to track it down and fix it :)

I'll go ahead and address the comment you made about the typeof vs iskind, but I don't know that I have the experience to look at the other issues. Is that okay with you? If not, I can close this PR out and just leave the linked issue open.

GrahamTheCoder:
Glad it was helpful. With the changes you've now made, I'm happy that this improves the converter sufficiently to merge. In a quick test I couldn't find the possibly related issue I mentioned, so I'll look at that separately another time.

Thanks!

PR #156, opened 2018-07-17
VB -> C#: Fix Qualify function to handle names that are already partially qualified

CoderNate:
This pull request contains an updated test for qualifying a name that is already partially qualified as well as a code change that makes the test pass.  I'm not sure if this is the correct way to handle it.  Also, there could be some scenarios with nested classes that this doesn't handle right.


GrahamTheCoder:
Great stuff. There was already a NestedClass test, but I added another one without redundant qualifiers which also passed fine. I'll close the related issue - massive thanks for all the help!

PR #188, opened 2018-10-19
Added ability to convert explicitly a solution or projects with no solution file

ceztko:
### Problem
The SolutionConverter offers a limited API to convert a single project that may have no solution linked.

### Solution
I added some convenience API to convert projects with no solution file or directly a `Solution`. This required some changes in the private functions to handle nullity of members like _solutionFilePath  or _projectReferenceReplacements. I briefly tested with both use cases, but I didn't run your tests: I'm beginning experimenting with your code to use it for completely different purpose, so feel free to accept this pull request or not.

### DIFF
--- a/ICSharpCode.CodeConverter/Shared/ProjectConversion.cs
+++ b/ICSharpCode.CodeConverter/Shared/ProjectConversion.cs
@@ -75,8 +75,10 @@ namespace ICSharpCode.CodeConverter.Shared
             var solutionFilePath = project.Solution.FilePath;
             var solutionDir = Path.GetDirectoryName(solutionFilePath);
             var compilation = project.GetCompilationAsync().GetAwaiter().GetResult();
-            var projectConversion = new ProjectConversion(compilation, compilation.SyntaxTrees.Where(t => t.FilePath.StartsWith(solutionDir)), languageConversion, GetConvertedCompilationWithProjectReferences(project, languageConversion));
-            foreach (var conversionResult in ConvertProjectContents(projectConversion)) yield return conversionResult;
+            var projectConversion = new ProjectConversion(compilation,
+                solutionDir == null ? compilation.SyntaxTrees : compilation.SyntaxTrees.Where(t => t.FilePath.StartsWith(solutionDir)),
+                languageConversion, GetConvertedCompilationWithProjectReferences(project, languageConversion));
+            return ConvertProjectContents(projectConversion);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
Thanks for your interest and suggestion. I'm a bit confused so let me start by explaining what was going on prior to this PR:
* If you want to convert a project contents, call `ProjectConversion.ConvertProjectContents`. 
* The `SolutionConverter` is for converting solutions. If you don't have one, just don't use it.
* There is a flaw in this separation: the `SolutionConverter` is currently the thing that can convert the project file itself.

It's not immediately obvious to me that this PR addresses the flaw, but was that your intention perhaps? In general I'd strongly advise against the pattern of assigning null to lots of things then trying to check them. I'd prefer to address the flaw is to either separate project file conversion into its own `ProjectFileConverter` class, or move it into `ProjectConversion`.

I'm interested to know (if you can share) the completely different purpose you're using the code for. The intent of the library is to be used for many different purposes, so I would like to merge changes to help that wherever possible.

GrahamTheCoder:
I've made some changes to make the API match my original intentions. Please feel free to add any test cases in another PR - hopefully my changes will work for your use case.

ceztko:
Thanks for taking your time to examine this pull request. Actually, I implemented the change in `SolutionConverter` because I dind't see a consistent API in `ProjectConversion`. It would be beautiful at this point if `ProjectConversion` had a `CreateFor<TLanguageConversion>(Project project)`, similar to `SolutionConverter.CreateFor(IReadOnlyCollection<Project> projects)`. I think you could also consider having a `SolutionConverter.CreateFor(Solution solution)` overload and modify the `IReadOnlyCollection<Project>` overload to take `IEnumerable<Project>` instead, as I did in the patch. As said, these are just suggestions. Actually, I will try to use your code as a learning boost for roslyn and try to create a C# to Java converter. It won't try to be a general purpose converter but it should be able to to convert code with some imposed constraints. I already understood that roslyn isn't very helpful for this purpose since I can't construct/inherit `SyntaxNode`.

GrahamTheCoder:
Ah yes, I agree the API isn't very consistent. Especially between static or non, and generic or not. I'm a big fan of read only collections, but it's true that at API boundaries I should take the most general parameters possible. Thanks for the feedback!

In converting to Java, the crucial thing you can use from here is the visitor structure that's already in place and shows how to delve into language details from the model. But you're right that what you return will need to be a java model which of course Roslyn doesn't have.
If it's a quick project just for fun you can just use a very simple type that basically contains a string and concat them together at the end. If you want something more serious, you can either find an existing java syntax tree model online, or create your own by subclassing syntax node and following the same pattern as the c# stuff. You can probably get a head start by cloning the Roslyn repo, copying the c# model then tweaking to fit java. Hope that helps

PR #190, opened 2018-10-25
Fix "if" statements

agametov:
### Problem

**CWE-561**. There are two '`if`' statements with identical conditional expressions. The first 'if' statement contains method return. This means that the second 'if' statement is senseless.

### Solution

Replace `SingleLineCommentTrivia `-> `SingleLineDocumentationCommentTrivia`

---
And small optimization: Replace `.Count()` -> `.Length`   _O(n) vs O(1)_


GrahamTheCoder:
Thanks for the pr. Looks right to me. If you could copy one of the member tests and tweak it to cover this case I'll merge it.

For your own knowledge, IIRC, in terms of algorithmic complexity there's no need to worry about the performance of `.Count()`. Its implementation is a little smarter than you might guess. It checks the passed object's type for known collection types and uses the appropriate property where available.
However, using the property directly will be a constant time faster, and it avoids other readers worrying about this, so I still think it's a sensible change.

Let me know if you need any help with the test

GrahamTheCoder:
Ah, I've just realized this is in the C# -> VB conversion which isn't currently used due to #32 not being finished. I'll merge as is and the test can be part of that PR.

PR #198, opened 2018-11-06
Fixes #197

### LINKED ISSUE #197: VB -> C#: WriteOnly Interface Properties Are Converted Incorrectly

### Input code
```VB
Interface TestInterface
    WriteOnly Property Items As Integer()
End Interface
```

### Erroneous output
```C#
interface TestInterface
{
    int[] Items { get; set; }
}
```

### Expected output
```C#
interface TestInterface
{
    int[] Items { set; }
}
```

### Details
Product in use: both

Version in use: e.g. 6.1.0

Currently working on creating a pull request for this one.

knodel12:
[Fixes #197 ](https://github.com/icsharpcode/CodeConverter/issues/197)

### Problem
WriteOnly properties on interfaces are converted incorrectly.

### Solution
* Just made a minor modification to the property syntax converter to watch for the WriteOnly keyword.

* [X] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
+++ b/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
@@ -263,17 +263,21 @@ namespace ICSharpCode.CodeConverter.CSharp
         public SyntaxToken ConvertIdentifier(SyntaxToken id, bool isAttribute = false)
         {
             string text = id.ValueText;
+
             var keywordKind = SyntaxFacts.GetKeywordKind(text);
             if (keywordKind != Microsoft.CodeAnalysis.CSharp.SyntaxKind.None)
                 return SyntaxFactory.Identifier("@" + text);
+
             if (id.SyntaxTree == _semanticModel.SyntaxTree) {
                 var symbol = _semanticModel.GetSymbolInfo(id.Parent).Symbol;
-                if (symbol != null && !String.IsNullOrWhiteSpace(symbol.Name)) {
+                if (symbol != null && !string.IsNullOrWhiteSpace(symbol.Name)) {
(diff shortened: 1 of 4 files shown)


ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
```diff
@@ -263,17 +263,21 @@ private string ConvertNumericLiteralValueText(string valueText, object value)
...
+                if (symbol != null && !string.IsNullOrWhiteSpace(symbol.Name)) {
                     if (symbol.IsConstructor() && isAttribute) {
                         text = symbol.ContainingType.Name;
                         if (text.EndsWith("Attribute", StringComparison.Ordinal))
                             text = text.Remove(text.Length - "Attribute".Length);
-                    } else if (symbol.IsKind(SymbolKind.Parameter) && symbol.Name == "Value") {
+                    } else if (symbol.IsKind(SymbolKind.Parameter) && symbol.ContainingSymbol.IsAccessorPropertySet() && ((symbol.IsImplicitlyDeclared && symbol.Name == "Value") || symbol.ContainingSymbol.GetParameters().FirstOrDefault(x => !x.IsImplicitlyDeclared) == symbol)) {
+                        // The case above is basically that if the symbol is a parameter, and the corresponding definition is a property set definition 
+                        // AND the first explicitly declared parameter is this symbol, we need to replace it with value.
                         text = "value";
```
GrahamTheCoder: Yep this is a great place to deal with this case. With code comments, I prefer to try to add context that isn't in the code like "override urges of a set parameter because c# always uses value, but vb can be anything" Or in many cases I'd pull it into a method or methods with a name that describes that so the comment isn't needed.
Thanks for the contribution!

PR #199, opened 2018-11-08
Fix array values assignment when there is a containing parent.

fabienDaou:
### Problem
**VB:**
parentClass.SomeArray(2) = "sometext"
**C# (before fix):**
parentClass(2) = "sometext"

I expect: parentClass.SomeArray[2] = "sometext"


GrahamTheCoder:
Looks good. I suspect there might be a more general case here that's still wrong, and I'm not sure about the leading trivia thing. I'll have time to take a closer look this weekend hopefully.

GrahamTheCoder:
I've added https://github.com/icsharpcode/CodeConverter/commit/231497e7b38ca3abd11a67d6a9bb8b0a4e1b3d0d and merged.

The thing that made me suspicious there was a more general case, was the use of HasLeadingTrivia. Trivia refers to comments, spaces, etc., but that didn't seem relevant to the overall test case.

Thanks for the contribution, sorry it a took a while to get merged

fabienDaou:
Thank you!

PR #204, opened 2018-11-12
Fix byref parameter

### LINKED ISSUE #264: VB -> C#: ref must be an assignable variable
### Input code
```vbnet
Public Class Class1
    Sub Foo()
        Bar(New Class1)
    End Sub

    Sub Bar(ByRef class1)
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    public void Foo()
    {
        Bar(ref new Class1());
    }

    public void Bar(ref object class1)
    {
    }
}

```

### Expected output
```c#
public class Class1
{
    public void Foo()
    {
        var class1 = new Class1();
        Bar(ref class1);
    }

    public void Bar(ref object class1)
    {
    }
}

```

### Details
Product in use:  VS extension

Version in use: 6.6.0

VB allows passing ref values directly without first assigning to a variable.

fabienDaou:
Fixes #264

### Problem
There is an issue with ByRef parameter that accept literal values when the method is invoked.

`DoSomething(ref "sometext")` is not supported in C# but it does not complain in VB (at least in my project.)

### Solution
For every literal values, I declared a local variable and used it instead.

Hope it helps,

Fabien

PS: I am not familiar with the framework to do code analyses so bear with me while I learn :)

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/MethodBodyVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/MethodBodyVisitor.cs
@@ -685,15 +685,15 @@ namespace ICSharpCode.CodeConverter.CSharp
             {
                 var invocationExpression = expression as InvocationExpressionSyntax;
                 if (invocationExpression != null) {
-                    InvocationExpressionSyntax ies;
-                    var list = AddRefVariables(invocationExpression, out ies);
-                    list.Add(SyntaxFactory.ExpressionStatement(ies));
-                    return SyntaxFactory.List(list);
+                    InvocationExpressionSyntax newInvocationExpression;
+                    var statements = GetNewStatements(invocationExpression, out newInvocationExpression);
+                    statements.Add(SyntaxFactory.ExpressionStatement(newInvocationExpression));
+                    return SyntaxFactory.List(statements);
                 }
                 return SyntaxFactory.SingletonList<StatementSyntax>(SyntaxFactory.ExpressionStatement(expression));
             }
 
-            private static List<StatementSyntax> AddRefVariables(InvocationExpressionSyntax initialInvocationExpression, out InvocationExpressionSyntax newInvocationExpression)
+            private static List<StatementSyntax> GetNewStatements(InvocationExpressionSyntax initialInvocationExpression, out InvocationExpressionSyntax newInvocationExpression)
             {
                 var newDeclarations = new List<StatementSyntax>();
                 var newArguments = new List<ArgumentSyntax>();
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
Thanks for the PR, you're definitely thinking along the right lines. 

In general, the rest of the code in the two main visitor classes avoid post-processing. I've added a point to the [contributing doc](https://github.com/icsharpcode/CodeConverter/blob/master/.github/CONTRIBUTING.md#implementation-advice) to help others in future avoid this pattern hopefully.

Before diving into changing that though, there's something else that I think's worth some thought. I notice that this approach won't work for ref literal in a field declaration, but whatever approach is used for a literal in a field declaration will probably work for this too. If that case seems too complex, I'd still merge a solution just to this problem within methods, since it makes things better, but let's throw out some ideas first.

e.g. Should we create a new method to contain the extra variable declarations, then call it with inline declarations as shown here where `DoSomethingOut` is generated and `DoSomethingElse` is the original callsite
```csharp
    public void DoSomethingElse()
    {
        DoSomethingOut1(out var val0, out var val1, out var dble, "sometext2");
    }

    public void DoSomethingOut1(out string str, out bool val, out double dble, string str2)
    {
        double dble = 2;
        var val0 = "sometext";
        var val1 = true;
        DoSomething(ref val0, ref val1, ref dble, str2);
    }
```
I think I'm right in saying this will work for all cases. In terms of implementation, we'd keep a list of these extra method declarations - see `NodesVisitor.AdditionalStatements`, and add them at the end, meaning the rest of the transformation could be done directly.

Let me know if you're interested in having a try at implementing this, or have another simpler idea.

Also I'm very happy to take feedback or contributions to help others get started by the way, I know it can be daunting.

Thanks again!

fabienDaou:
I will change the code to avoid post-processing as suggested.

Just to be sure I understand. By "ref literal in a field declaration", do you mean this:
`Class Surrounding
	Public Str As String = DoStuff("bla")
	
	Private Shared Function DoStuff(ByRef str As String) As String
		str = "DoStuff"
		Return str
	End Function
End Class`

By the way, I am not going to be able to contribute for the next two weeks so it might take some times before I make the changes.

GrahamTheCoder:
Yep that's the one. OK thanks for the heads up - I'll leave this open. Remember to branch from, merge or rebase onto master when you continue, and feel free to totally nuke my commits if they're no longer relevant of course.

GrahamTheCoder:
Are you still interested in continuing this @fabienDaou ?

fabienDaou:
Kind of ghosted github recently... I see there are changes made to fix this issue, I also tested it on my own code and it seemed to work fine.

Thanks for the fix!

Issue #184, opened 2018-10-10
Solution/project tests broken in VS 15.8.6+


GrahamTheCoder:
See the project and solution tests. No cs or vb files get generated because no syntax trees are found in the loaded projects. 

This is almost certainly due to the changes to separate MSBuild from Visual Studio:
* https://github.com/dotnet/roslyn/issues/21583
* https://github.com/dotnet/roslyn/issues/21644
* https://github.com/neuecc/MessagePack-CSharp/issues/108

It doesn't affect nuget/web/vsix, but could affect other consumers who make use of MSBuild, so it may still get reported here.

taylus:
It took me a while to get context on this, but once I did I was able to repro it nicely. I had an older version of VS2017 and updating it broke it like you said where the [project's syntax trees](https://github.com/icsharpcode/CodeConverter/blob/5b2749b8fb4014d802043fafb60b51699ec48cac/ICSharpCode.CodeConverter/Shared/ProjectConversion.cs#L88) were always an empty collection.

I was able to fix this in a throwaway console app by installing/updating the Microsoft.Build packages referenced in dotnet/roslyn#21644 to their latest versions (v15.9.20). Doing this in CodeConverter gets you the syntax trees back, but the [test I've been running](https://github.com/icsharpcode/CodeConverter/blob/5b2749b8fb4014d802043fafb60b51699ec48cac/Tests/CSharp/SolutionAndProjectTests.cs#L22) is failing on a difference in the Application.Designer.vb -> .cs file. The C# version is an empty namespace but the VB version is just some options which convert to nothing.

I was seeing this before I updated VS too, so it might be something that crept in while these tests were skipped. Any ideas?

![image](https://user-images.githubusercontent.com/8031544/49127407-c9377c00-f28c-11e8-8837-d4448844ad4e.png)

taylus:
Here is a WIP: https://github.com/taylus/CodeConverter/tree/184-msbuild-update

taylus:
Ok, I'll probably just tweak the actual or expected since an empty namespace declaration isn't much different from an empty file. I'll take another look tonight and then loop back around to your ideas in #210

GrahamTheCoder:
Yes for now setting the expected to the actual is fine. These tests were intended to characterize the current behavior rather than being a perfect conversion. When reviewing the PR we should raise issues for any changes which look like regressions.

taylus:
[...]

Issue #60, opened 2018-03-20
Enable checks that test conversion code compiles


GrahamTheCoder:
Originally this was to track a collection of test improvements, the title has been updated to reflect remaining task.

Either add new tests, or add to the current test harness things like:
* Test that the input and output of existing tests compiles - they're just text strings and it's easy to accidentally be verifying incorrect behaviour
* Once the above is done, it may at least be interesting diffing the IL compiled from the input and output (obviously won't match in all cases, but would give great confidence for the ones it works for)
* Roundtripping e.g. VB -> C# -> VB -> C# and test the results of the first and third conversion are the same (don't expect both VB to be the same in this instance due to whitespace etc.)
* Tests with self-verifying inputs - i.e. We can run the input and output and check they have the same result.

siegfriedpammer:
> Diffing the IL compiled from the input and output (obviously won't work in all cases, but would give great confidence for the ones it works for)
This won't work at all, except for very simple programs, because csc and vbc produce very different IL most of the time.

Suggestion: Testing Behavior? That is, convert a program, compile the source and the converted program, execute both and compare their output. If both compile and produce the same output, conversion is successful.

GrahamTheCoder:
That is what I thought of first, and indeed I've worked on something previously where actually running the input and output worked well so I'll take the recommendation and add it to the description.

I figured the IL being equivalent is the step up from knowing the program has the same result for one or two inputs. Maybe I'm overestimating how many of even the simple cases it'll work for, but I think it's really easy to implement so I'll probably try it anyway and see.

GrahamTheCoder:
The self verifying tests introduced in #216 allow the deeper level of testing this issue was intended to track and their use has been proven in #210

The remaining task to close this issue is to turn on reporting of compile errors in the normal tests (with some way to override it, since imperfect code should still be convertible)

GrahamTheCoder:
I recently added warnings by copying the VB namespace code in. These will need to be fixed as part of this

GrahamTheCoder:
Covered by #539

PR #210, opened 2018-11-15
No longer explicitly cast rhs of floating point division to double

### LINKED ISSUE #202: VB -> C# Decimal casted to double
### Input code
```vbnet
return 100D / 100D
```

### Erroneous output
```C#
{
    return 100M / (double)100M;
}
```

### Expected output
```C#
{
    return 100M / 100M;
}
```

### Details
Product in use: both

Version in use: e.g. 6.1

taylus:
Fixes #202 maybe


GrahamTheCoder:
My suspicion is that this will be wrong for integers. I'll do some testing.

taylus:
Yeah I'm actually not even sure about the original issue anymore. I put it into http://converter.telerik.com/ and get the same double cast.

GrahamTheCoder:
The only way to know for sue is to actually run the result. I didn't get around to it yet, let me know if you do.

I'm afraid the telerik converter is actually just a less frequently updated build of this converter with their branding and copyright notice added on, so not an independent data point.

taylus:
I ran some tests and you're right!

I feel like I'm missing something with the /= operator though.

VB.NET
``` vb
Dim x = 5 / 2
Console.WriteLine($"{x.GetType()}: {x}")    'prints System.Double: 2.5

Dim y = 5
y /= 2
Console.WriteLine($"{y.GetType()}: {y}")    'prints System.Int32: 2 (?!)
```

C#
``` csharp
var x = 5 / 2;
Console.WriteLine($"{x.GetType()}: {x}");    //prints System.Int32: 2

var y = 5;
y /= 2;
Console.WriteLine($"{y.GetType()}: {y}");    //prints System.Int32: 2
```

GrahamTheCoder:
I imagine since /= knows the answer is an int (because the variable is if type int), it does int division. But yeah it can be hard to pin down these things. For this sort of case I'd like to write a test that compiles and runs both bits and assets they have the same result, but I haven't got around to writing a helper function to do that yet

taylus:
Hey @GrahamTheCoder, your idea of a test that compiles and runs the source intrigued me and I started [playing with the idea here](https://github.com/taylus/roslyn).

Would you like me to see if I can incorporate that here? I've never used Roslyn before so it was a learning experience and I'm not sure if I'm doing things right (my `ReferenceResolver` class feels weird but "it works on my machine" :stuck_out_tongue:).

GrahamTheCoder:
Looks good. Before diving into implementation, I want to think a bit about the most maintainable structure, in terms of number of files, and what we assert on (stfout?) This relates a bit to #60 btw

taylus:
The console output redirection does feel awkward and I bet will cause problems for tests running in parallel. Outside of returning different status codes from Main() though I'm not sure what else you could do. I wonder if System.Diagnostics.Process would work (better?) for this.

GrahamTheCoder:
I'm looking into doing something a bit lower than assembly level using `CSharpScriptEngine`
https://blog.jayway.com/2015/05/09/using-roslyn-to-build-a-simple-c-interactive-script-engine/
I'll update here after
Though first I'm going to take a brief digression to update the csproj format to avoid `FileNotFoundException`s caused by binding redirect issues

GrahamTheCoder:
My digression was bigger than expected, so I've put it aside and found a workaround here.
As it turns out the there is no VB equivalent of the scripting namespace (at this time):
https://github.com/dotnet/roslyn/issues/13523

I could imagine using an approach like the one you prototyped if we took it a bit further and used reflection to call a known test method with a carefully built up string output (to give more flexibility and avoid the tie to stdout). Would you be interested in trying to fix #184 which does on disk conversion, then use code like in your prototype to compile them both and run the methods within?

## Option 1
I think if we can get XUnit to treat each method in the compiled assemblies as a separate test if we use one of these [MemberData](https://stackoverflow.com/a/51172545/1128762) attributes to enumerate methods (use reflection to look for a "ExecutableTestMethod" attribute) then match them up between the source and destination compiled assembly.

## Option 2
What I've described above is close to what a test assembly already looks like. We could write the tests in such a way that they self verify and just run the normal test runner on them. `Assert.Equal(2 / 2,  1);` etc. 
I have a hunch that getting the test runner to do the right thing in local builds and on the build server, ensuring the projects have been converted and written to disk first will be slightly annoying. But it does have a certain simplicity to it.

## Option 3
If you combine the above two approaches using the xunit sdk or some such to programmatically run the test runner, and turn those into test results in the current session.

Let me know if you want more info, or just to discuss a bit more.

taylus:
Sure, I will take a look and see if I can help!

GrahamTheCoder:
Yes I think taylus/roslyn@c1b5c18 looks like one of the options I was thinking of (sorry they weren't very clearly delineated). Honestly I'm not sure there are the original options I wrote make that much sense! I think the details of how to integrate the test results neatly are mostly orthogonal from the approach to writing the tests, so here are two new options regarding writing the tests.

### Test writing approach A
Write self-verifying tests.

The test process:
* Run it in its original form as a sanity check that it's a valid test case.
* Convert it
* Run it in its converted form to check it still passes

i.e. The test case we'd commit would be
```
<Fact>
Public Sub TestDivision
    Dim x = 7 / 2
    Assert.Equal(x, 3.5)
End Sub
```

Pros:
* Don't need to write or commit the expected conversion
Cons:
* There are some ways the conversion could mess up that'd still let the test pass:
  * Totally omitting/distorting assertions
  * Two bugs that happen to cancel out

### Test writing approach B
```
<OutputVerifiedTest("3.5")>
Public Function TestDivision
    Dim x = 7 / 2
    Return x.ToString()
End Function
```

The test process:
* Read the expected value from the attribute
* Run it in its original form as a sanity check against that expected value.
* Convert it
* Run it in its converted form and check it matches the expected value.

Pros: 
* Don't need to write or commit the expected conversion
* Since we read the expected output from the source file, and test both against it, it slightly reduces the possibility of a conversion error in converting the expected value.

Cons:
* Less flexibility in assertions (tied to strings)
* There are some ways the conversion could mess up that'd still let the test pass:
  * Two bugs that happen to cancel out

### Comparison

I think making use of existing assertions library from xUnit is of sufficient benefit to choose A.
Can you see any other alternatives to consider?

Other notes:
* You do have to get the initial test right, but it'll be easy to find by running it, and seeing what the assertion failure message is at the sanity check phase.
* If the test fails, it'd need to put the conversion input/output in the test output.
* We could (on an opt-out per-test basis perhaps) "round trip" the conversion - and check it still passes when converted back again to the original language.
* We should use [xUnit MemberData](https://stackoverflow.com/a/51172545/1128762) to populate each as an individual test case in the results

taylus:
I agree and think approach A is the better one too. 

(I was thinking yesterday of something like approach B but the method would return x instead of its string representation and the test would compare it before and after conversion. I'm not sure how well that would work though and would have to experiment.)

I will see what I can do with the xUnit approach!

taylus:
I think this might be a snag. If I do something like this:

```csharp
[Theory, MemberData(nameof(TestData))]
public void ParameterizedTest(object expected, object actual)
{
    Assert.Equal(expected, actual);
}

public static IEnumerable<object[]> TestData => new List<object[]>
{
    new object[] { 7 / 2, 3 },
    new object[] { 7 / 2.0, 3.5 }
};
```

That compiles down to testing that 3 == 3 and 3.5 == 3.5 by the time reflection gets at the test data. Are we testing the compiler at this point and not the converter?

GrahamTheCoder:
```csharp


    public class SelfVerifyingTests
    {
        [MemberData(nameof(VisualBasicToCsharpTestData))]
        public void VisualBasicToCsharp(Action verifyConvertedTestPasses)
        {
            verifyConvertedTestPasses();
        }

        public static IEnumerable<object[]> VisualBasicToCsharpTestData {
            get {
                // TODO error handling around each of these parts so it's clear what stage went wrong
                var sourceFileText = "../../TestData/CSharp/SelfVerifyingTest.vb";
                var compiledSource = CompilerFrontend.FromFile(sourceFileText, "SourceAssembly").Compile();
                var convertedText =
                    ProjectConversion.ConvertText<VBToCSConversion>(sourceFileText, DefaultReferences.NetStandard2);
                var compiledTarget = CompilerFrontend.FromFile(convertedText, "TargetAssembly").Compile();
                Dictionary<string, Action> runnableTestsInSource = TestRunner.GetSourceTests(compiledSource);
                Dictionary<string, Action> runnableTestsInTarget = TestRunner.GetSourceTests(compiledTarget);
                List<Action> selfVerifyingTests = runnableTestsInSource.Select(sourceTest =>
                    () => {
                        sourceTest.Value();
                        runnableTestsInTarget[sourceTest.Key]();
                    }
                ).ToList();
                return selfVerifyingTests;
            }
        }
    }
```

Obviously you'll need to tweak your current test runner to return the fully qualified name of the test and an action which executes it, rather than executing it directly.

Hope that helps!

GrahamTheCoder:
Now that #216 is merged, we can get back to the matter at hand 😃. Do you want to rebase onto master and see if you can get the commented out test passing?

taylus:
Sure! I think I did it right.

TestData/SelfVerifyingTests/VBToCS/ExpressionTests.vb
```diff
@@ -7,30 +7,37 @@ Module Program
...
+            Assert.Equal(x, 3.5D)
+        End Sub
+        
+        <Fact>
+        Public Sub TestIntegerDivisionOfChars()
+            Dim x As Char = 7
+            Dim y As Char = 2
+            Dim z = x / y
+            Assert.Equal(x, 3)
+        End Sub
```
GrahamTheCoder: This turned out not to be a thing - my Visual Studio was playing up so I couldn't try it out

GrahamTheCoder:
Great stuff, thanks! If you're interested in doing more, feel free to pick up pretty much any of the existing issues. There are two that would particularly benefit from this new style of test:
https://github.com/icsharpcode/CodeConverter/issues?utf8=%E2%9C%93&q=is%3Aopen+label%3A%22output+logic+error%22+label%3A%22VB+-%3E+C%23%22+

PR #216, opened 2018-11-29
Self verifying executable tests

taylus:
WIP PR for #184


GrahamTheCoder:
I've added a change needed to compile on my machine.
I've also added the characterization that passes on my machine. Could you check whether it works for you locally? It's a slightly odd issue where the System reference can't be resolved, so if it's working differently on your machine we'll need either more investigation.

taylus:
:D

Those changes actually cause the VB -> C# solution and project tests to fail on my end. The "Attribute" name doesn't appear in the attribute when it converts. Weird -- I wonder what the build server will do.

image

Tests/CSharp/SelfVerifyingTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            } catch (CompilationException ex) {
+                throw new XunitException($"Error compiling target: {ex}");
+            }
+        }
+
+        private static IEnumerable<MetadataReference> AdditionalReferences => new List<MetadataReference>()
+        {
+            MetadataReference.CreateFromFile(typeof(Assert).Assembly.Location),
+            MetadataReference.CreateFromFile(typeof(FactAttribute).Assembly.Location),
+            MetadataReference.CreateFromFile(Assembly.Load("System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a").Location), //this feels fragile and wrong but I'm not sure
```
taylus: Iffy about how these are referenced but couldn't find a better way.

taylus:
Hey @GrahamTheCoder, I think I better understand your latest comments in #210 and pushed some more changes. What do you think? I might be duplicating some compilation code.

GrahamTheCoder:
^ I've rebased it onto master and removed my previous commit - after a git clean, my machine, yours and the build server all agree on the conversion of those attributes.

It's looking really good - I'll make some changes on top and get it merged if possible

taylus:
Np! Thank you for cleaning it up!

PR #273, opened 2019-04-08
Add initial support for VB MyClass expressions [#260]

### LINKED ISSUE #260: VB -> C#: Error converting shared/static variable

### Input code
```vbnet
Public Class Class1
    Sub Foo()
        MyClass.Val = 6
    End Sub

    Shared Val As Integer
End Class
```

### Erroneous output
```c#
public class Class1
{
    public void Foo()
    {
        ;/* Cannot convert AssignmentStatementSyntax, CONVERSION ERROR: Conversion for MyClassExpression not implemented, please report this issue in 'MyClass' at character 46
   at ICSharpCode.CodeConverter.CSharp.VisualBasicConverter.NodesVisitor.DefaultVisit(SyntaxNode node)
   at Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxVisitor`1.VisitMyClassExpression(MyClassExpressionSyntax node)
   at Microsoft.CodeAnalysis.VisualBasic.Syntax.MyClassExpressionSyntax.Accept[TResult](VisualBasicSyntaxVisitor`1 visitor)
   at Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxVisitor`1.Visit(SyntaxNode node)
   at ICSharpCode.CodeConverter.CSharp.CommentConvertingNodesVisitor.DefaultVisit(SyntaxNode node)
   at Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxVisitor`1.VisitMyClassExpression(MyClassExpressionSyntax node)
   at Microsoft.CodeAnalysis.VisualBasic.Syntax.MyClassExpressionSyntax.Accept[TResult](VisualBasicSyntaxVisitor`1 visitor)
   at ICSharpCode.CodeConverter.CSharp.VisualBasicConverter.NodesVisitor.VisitMemberAccessExpression(MemberAccessExpressionSyntax node)
   at Microsoft.CodeAnalysis.VisualBasic.Syntax.MemberAccessExpressionSyntax.Accept[TResult](VisualBasicSyntaxVisitor`1 visitor)
   at Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxVisitor`1.Visit(SyntaxNode node)
   at ICSharpCode.CodeConverter.CSharp.CommentConvertingNodesVisitor.DefaultVisit(SyntaxNode node)
   at Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxVisitor`1.VisitMemberAccessExpression(MemberAccessExpressionSyntax node)
   at Microsoft.CodeAnalysis.VisualBasic.Syntax.MemberAccessExpressionSyntax.Accept[TResult](VisualBasicSyntaxVisitor`1 visitor)
   at ICSharpCode.CodeConverter.CSharp.VisualBasicConverter.MethodBodyVisitor.VisitAssignmentStatement(AssignmentStatementSyntax node)
   at Microsoft.CodeAnalysis.VisualBasic.Syntax.AssignmentStatementSyntax.Accept[TResult](VisualBasicSyntaxVisitor`1 visitor)
   at Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxVisitor`1.Visit(SyntaxNode node)
   at ICSharpCode.CodeConverter.CSharp.CommentConvertingMethodBodyVisitor.ConvertWithTrivia(SyntaxNode node)
   at ICSharpCode.CodeConverter.CSharp.CommentConvertingMethodBodyVisitor.DefaultVisit(SyntaxNode node)

Input: 
        MyClass.Val = 6

 */
    }

    private static int Val;
}

```

### Expected output
```vbnet
public class Class1
{
    public void Foo()
    {
        Class1.Val = 6;
    }

    private static int Val;
}

```

### Details

Version in use: e.g. 6.6.0

mrmonday:
Closes #260

This should implement the majority of `MyClass` support for VB -> C#. There is one case (that I'm aware of) that is missing - that should throw an exception.

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -1390,7 +1390,7 @@ namespace ICSharpCode.CodeConverter.CSharp
                 return _queryConverter.ConvertClauses(node.Clauses);
             }
 
-            private SyntaxToken ConvertIdentifier(SyntaxToken identifierIdentifier, bool isAttribute = false, bool updateCase = false)
+            private SyntaxToken ConvertIdentifier(SyntaxToken identifierIdentifier, bool isAttribute = false, bool updateCase = true)
             {
                 return CommonConversions.ConvertIdentifier(identifierIdentifier, isAttribute, updateCase);
             }
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -454,6 +454,7 @@ namespace ICSharpCode.CodeConverter.CSharp
                 var convertibleModifiers = node.Modifiers.Where(m => !m.IsKind(VBasic.SyntaxKind.ReadOnlyKeyword, VBasic.SyntaxKind.WriteOnlyKeyword, VBasic.SyntaxKind.DefaultKeyword));
                 var modifiers = CommonConversions.ConvertModifiers(convertibleModifiers, GetMemberContext(node));
                 var isIndexer = node.Modifiers.Any(m => SyntaxTokenExtensions.IsKind(m, VBasic.SyntaxKind.DefaultKeyword));
+                var accessedThroughMyClass = IsAccessedThroughMyClass(node, node.Identifier, _semanticModel.GetDeclaredSymbol(node));
 
                 var initializer = (EqualsValueClauseSyntax)node.Initializer?.Accept(TriviaConvertingVisitor);
                 var rawType = (TypeSyntax)node.AsClause?.TypeSwitch(
@@ -486,7 +487,12 @@ namespace ICSharpCode.CodeConverter.CSharp
                     );
                 }
 
-                if (isIndexer)
+                if (isIndexer) {
+                    if (accessedThroughMyClass) {
+                        // TODO Not sure if this is possible?
+                        throw new NotImplementedException("MyClass indexing not implemented");
+                    }
+
                     return SyntaxFactory.IndexerDeclaration(
                         SyntaxFactory.List(attributes),
                         modifiers,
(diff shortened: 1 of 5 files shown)


GrahamTheCoder:
I've rebased to fix the line endings

GrahamTheCoder:
For naming, I'd go for MyClassF1.
In the place of a to-do it'd be fine to just throw. It should already get caught at the statement level and turned into a conversion error comment
(I've just had a thought, maybe rather than comments they should be #error or #warning directives but that's a separate issue)

When converting a virtual method, look for usages of its symbol within the class (there's a Roslyn API for finding usages) and see if any are a myclass reference (not quite sure how to do the second bit, but it has to be symbolic to deal with partial classes).
If found, add the normal result for the visit method to the "additional declarations" field and just return a method call to the new method.

Let me know if that doesn't make enough sense and I'll go and poke around the code a bit.

mrmonday:
I have a more complete version of this now, I'll hopefully update the PR later today.

mrmonday:
I'm pretty happy with where this is now - there's one case missing, which I've documented, and will throw an exception.

Finding if we need to introduce an extra method is a little messy - I traverse up the AST to find the class we're in, then back down again. I looked into using `SymbolFinder`, but that requires a `Solution` object, which we don't seem to have available. We also don't need to find *all* references, only those in the same class. We could potentially skip this, and generate the extra method for every virtual call, that's quite ugly though.

There are a few things which could probably be cleaned up, happy do do so if there's anything specific that you see.

Hopefully I've not messed up line endings again here.

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -571,6 +601,24 @@ private BlockSyntax VisitStatements(SyntaxList<VBSyntax.StatementSyntax> stateme
                 return SyntaxFactory.Block(statements.SelectMany(s => s.Accept(methodBodyVisitor)));
             }
 
+            private static bool IsAccessedThroughMyClass(SyntaxNode node, SyntaxToken identifier, ISymbol symbol)
+            {
+                bool accessedThroughMyClass = false;
+                if (symbol.IsVirtual && !symbol.IsAbstract) {
+                    var classBlock = node.Ancestors().OfType<VBSyntax.ClassBlockSyntax>().SingleOrDefault();
```
mrmonday: This needs to be `FirstOrDefault()` in case there are nested classes. Need to double check the order too.

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -495,13 +501,37 @@ public override CSharpSyntaxNode VisitPropertyStatement(VBSyntax.PropertyStateme
...
                         accessors
                     );
-                else {
+                } else {
+                    var identifier = node.Identifier;
+                    if (accessedThroughMyClass) {
+                        var identifierName = "MyClass" + node.Identifier.ValueText;
+                        var getReturn = SyntaxFactory.Block(SyntaxFactory.ParseStatement($"return this.{identifierName};"));
+                        var getAccessor = SyntaxFactory.AccessorDeclaration(SyntaxKind.GetAccessorDeclaration, getReturn);
+                        var setValue = SyntaxFactory.Block(SyntaxFactory.ParseStatement($"this.{identifierName} = value;"));
```
GrahamTheCoder: I'll add a call to ConvertIdentifier first and change the test to include some inconsistent VB casing.

Especially when working with strings, its really easy to lose track of what language they're in. You'll see I've started prefixing some variable names with `cs` or `vb` to help with this in complicated areas. I feel like the `SyntaxFactory.Parse*`  methods make it easier to get lost due to encouraging more strings. I think you've used them in moderation at a sensible level in this case, so good job!

GrahamTheCoder:
Ah I'm just going to check what happens when there are multiple references to the same property via MyClass too.

GrahamTheCoder:
Since all the tests pass, I've enabled `updateCase` in all cases. If there was a situation you were trying to avoid by having it optional let me know.

Tests/CSharp/NamespaceLevelTests.cs
```diff
@@ -379,7 +379,7 @@ public void MyClassVirtualCallProperty()
     Overridable Property P1() As Integer = 1
     MustOverride Property P2() As Integer
     Public Sub TestMethod() 
-        Dim w = MyClass.P1
+        Dim w = MyClass.p1
```
mrmonday: Was this intentional? This doesn't look right.

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
mrmonday: I skipped `ConvertIdentifier`, since it could prefix the variable with `@`, which would break this. Perhaps it should concat then convert?

I'll make sure to prefix with `cs` and `vb` in future.

mrmonday:
`updateCase` was mostly to not break things - I hadn't tested everywhere it was called. If it works everywhere, that's great!

Tests/CSharp/NamespaceLevelTests.cs
GrahamTheCoder: The intent was to test that it still works even when it's been accessed with the "wrong" case in VB, I actually meant to add a comment to the test to say that - will do so

mrmonday: My other comment got lost somehow, realised that afterwards. Looks fine :)

PR #274, opened 2019-04-08
Re-order constraints when converting from VB to C#

### LINKED ISSUE #271: VB -> C#: New constraint in wrong position
### Input code
```vbnet
Public Interface Foo
End Interface

Public Class Bar(Of x As {New, Foo})

End Class
```

### Erroneous output
```c#
public interface Foo
{
}

public class Bar<x> where x : new(), Foo
{
}
```

### Expected output
```c#
public interface Foo
{
}

public class Bar<x> where x : Foo, new()
{
}
```

### Details
Product in use: VS extension

Version in use: 6.6.0

Error from C#:
```
CS0401	The new() constraint must be the last constraint specified
```

mrmonday:
Closes #271 

### Problem

When converting from VB to C#, new constraints in type parameters need to be the last constraint.

### Solution

Added an OrderBy() to ensure that the new constraint ends up last. I haven't checked that the order of the other constraints remains the same, but I assume it does.

* [x] All tests pass


GrahamTheCoder:
First PR merged, thanks!

PR #275, opened 2019-04-08
Ensure base.Finalize() is removed

### LINKED ISSUE #269: VB -> C#: Do not explicitly call Finalize
### Input code
```vbnet
Public Class Class1
    Protected Overrides Sub Finalize()
        MyBase.Finalize()
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    ~Class1()
    {
        base.Finalize();
    }
}
```

### Expected output
```c#
public class Class1
{
    ~Class1()
    {
    }
}
```

### Details
Product in use: VS extension

Version in use: 6.6.0

Error from C#:

```
CS0250	Do not directly call your base class Finalize method. It is called automatically from your destructor.
```

mrmonday:
Closes #269 

### Problem

base.Finalize() is not removed when converting from VB to C#

### Solution

I'm not sure this is the right/nicest way of handing this:

 * I don't check that we're actually inside `~Finalize()`
 * It adds an overhead for every expression statement
 * I wasn't sure how to just add a comment, so there's a leading `;` in there. I decided a comment would be useful, since we're removing code - I'm happy to leave nothing there instead though.

* [ ] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/MethodBodyVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/MethodBodyVisitor.cs
@@ -105,6 +105,12 @@ namespace ICSharpCode.CodeConverter.CSharp
 
             public override SyntaxList<StatementSyntax> VisitExpressionStatement(VBSyntax.ExpressionStatementSyntax node)
             {
+                if (node.Expression is VBSyntax.InvocationExpressionSyntax invoke &&
+                    invoke.Expression is VBSyntax.MemberAccessExpressionSyntax expr &&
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
Thanks for the PR, looks good to me!
Yep, in this visitor pattern it is slightly annoying to just add a comment since it has to attach to something. I think just returning an empty list would be slightly preferable since it's a correct conversion and the comment might confuse people into trying to uncomment it. Would you mind switching it to do that before I merge it?

mrmonday:
Done!

PR #276, opened 2019-04-08
Automatically fix the casing of identifiers where possible

### LINKED ISSUE #258: VB -> C#: Fix casing

### Input code
```vbnet
Public Class Class1
    Sub Bar()

    End Sub

    Sub Foo()
        bar()
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    public void Bar()
    {
    }

    public void Foo()
    {
        bar();
    }
}
```

### Expected output
```vbnet
public class Class1
{
    public void Bar()
    {
    }

    public void Foo()
    {
        Bar();
    }
}
```

### Details
Product in use: VS extension

Version in use: 6.6.0

Visual Basic is case insensitive. When converting code to C#, it should fix the casing of method calls, properties, etc.

mrmonday:
Closes #258 

### Problem

VB allows calling methods using any casing/capitalization, where C# does not. Attempt to fix the casing where possible.

### Solution
* When converting identifiers, fix-up the casing
* This might be needed in other places too, but this fixes the original bug

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
+++ b/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
@@ -274,7 +274,7 @@ namespace ICSharpCode.CodeConverter.CSharp
             return value.ToString();
         }
 
-        public SyntaxToken ConvertIdentifier(SyntaxToken id, bool isAttribute = false)
+        public SyntaxToken ConvertIdentifier(SyntaxToken id, bool isAttribute = false, bool updateCase = false)
         {
             string text = id.ValueText;
 
@@ -285,6 +285,10 @@ namespace ICSharpCode.CodeConverter.CSharp
             if (id.SyntaxTree == _semanticModel.SyntaxTree) {
                 var symbol = _semanticModel.GetSymbolInfo(id.Parent).Symbol;
                 if (symbol != null && !string.IsNullOrWhiteSpace(symbol.Name)) {
+                    if (updateCase && text.Equals(symbol.Name, StringComparison.OrdinalIgnoreCase)) {
+                        text = symbol.Name;
+                    }
+
                     if (symbol.IsConstructor() && isAttribute) {
                         text = symbol.ContainingType.Name;
                         if (text.EndsWith("Attribute", StringComparison.OrdinalIgnoreCase))
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -1298,9 +1298,9 @@ namespace ICSharpCode.CodeConverter.CSharp
                 return _queryConverter.ConvertClauses(node.Clauses);
             }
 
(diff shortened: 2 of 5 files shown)


mrmonday:
Two DeclareStatement tests fail after this change - I am not sure why.

It seems the `using System.Diagnostics;` now get removed... I have no idea why that is though, or how to fix it - I'd appreciate some insight there.

GrahamTheCoder:
Wow, that certainly is an unexpected effect based on what you changed.
Since it's in the VB source, either something went wrong with the conversion and it got silently swallowed, or this code decided to treat it differently: https://github.com/icsharpcode/CodeConverter/blob/efe1294e3d79d25f6a2eac8fd2b3dcc8d15546dd/ICSharpCode.CodeConverter/CSharp/CompilationErrorFixer.cs#L60

The code to clean up usings happens in a second pass and only removes them if it finds no compilation errors missing types (in case it's actually a missing reference that the using statement would have correctly pointed to). I don't know how your changes could affect any of that since the rest of the code is identical though. Let me know if you're still struggling after sticking a breakpoing on the "using" conversion and on the "using" tidyup.

GrahamTheCoder:
Ah, I see it now, the original output didn't compile. You've correctly fixed the casing of process.GetProcesses()

The second pass must have been unable to do its thing until now. Feel free to update the expectation and I'll merge.

GrahamTheCoder:
Is the input code just ambiguous since it uses process with two different casings? If so, feel free to just change the variable name.

mrmonday:
OK, this is super weird - it seems like a Roslyn bug...

The `GetSymbolInfo()` for the `Process` in `Process.GetProcesses()` comes back with the local symbol `process` - but that shouldn't be in scope there. If it is, the whole loop shouldn't compile. Unless VB does some magic to try and guess which to use, but then we should get back either the correct symbol, or at the very least a list of candidates in the `SymbolInfo` (we don't).

mrmonday:
I think you might be right - it might not compile. Testing now.

mrmonday:
Gives one of these: https://docs.microsoft.com/en-us/dotnet/visual-basic/misc/bc30980

I'll fix up the test case.

mrmonday:
Fixed and rebased. Unfortunately my mergetool messed up the line endings, so the diff is messy... I'll see if I can fix it.

mrmonday:
I'm not sure what happened there, will leave it for now. Feel free to fix it if it bothers you.

GrahamTheCoder:
Ah yes, I hadn't realized the line endings were changed by your previous PRs. I've changed them back to CRLF in master. If you set `git config core.autocrlf true` for this repo, it should do the right thing from now on while everything's consistent.
I usually use notepad++'s `edit->EOL conversion` or a grep regex replace to sort things like this out depending how many files need changing.

mrmonday:
Sorry about that :(

It might be a good idea to add a `.gitattributes` file to force the line endings to be correct once they get as far as git.

GrahamTheCoder:
No worries. I'll see if I can figure out the .gitattributes thing at some point. I've had issues with it in the past, but I think that was in the case where the repo was already inconsistent

PR #282, opened 2019-04-17
Add #error directive on conversion error

mrmonday:
Link to issue(s) this covers

### Problem
When conversion fails, a comment is added, with code not converted. This is an incorrect translation and should cause a compilation error.

### Solution
This adds a `#error`  directive before the comment to make it obvious when conversion has failed.


GrahamTheCoder:
Looks goo to me - just added an explicit check for #error to the test for this

mrmonday:
I knew I'd seen a test somewhere for it, couldn't find it though! Thanks!

Issue #174, opened 2018-09-09
vb.net -> C#: Replace Visual Basic library calls with equivalent things from the .NET framework


B0R1K:
instead of using: 
`!Information.IsDBNull`

it should use 
`!DBNull.Value.Equals`

EDIT: See https://github.com/icsharpcode/CodeConverter/issues/174#issuecomment-461091117 for fuller listing

GrahamTheCoder:
Request to substitute Vb, Cr and VbCrlf with their string equivalent: https://github.com/icsharpcode/CodeConverter/issues/228

GrahamTheCoder:
Here are some that I came across when moving to things to .NET Standard:
* `Hex(byte)` to $"{byte:X2}"
* `My.Computer.FileSystem.DirectoryExists(fileName)` to `Directory.Exists(fileName)`
* `My.Computer.FileSystem.FileExists(fileName)` to `File.Exists(fileName)`
* `My.Computer.FileSystem.DeleteFile(fileName)` to `File.Delete(fileName)`
* `My.Computer.Info.OSFullName` to `Runtime.InteropServices.RuntimeInformation.OSDescription`
* `My.Computer.FileSystem.GetFileInfo(f)` to `new FileInfo(f)`
* `DateDiff(DateInterval.Day, aDate, aSecondDate)` to `(aDate - aSecondDate).TotalDays`
* `DateDiff(DateInterval.Hour, aDate, aSecondDate)` to `(aDate - aSecondDate).TotalHours`
* `DateDiff(DateInterval.Minutes, aDate, aSecondDate)` to `(aDate - aSecondDate).TotalMinutes`
* `My.Computer.FileSystem.SpecialDirectories.Desktop` to `Environment.GetFolderPath(Environment.SpecialFolder.Desktop)`
* `UBound(myArray)` to `myArray.GetUpperBound(0)`
* `Format$(value)` to `$"{value}"`
* `Format$(value, "0.0")` to `$"{value:0.0}"`
* `Replace(strOutPut, "&lt;", "<")` with `strOutPut.Replace("&lt;", "<")`
* `Len(str)` to `str.Length`
* `Left(x, 4)` to `x.Substring(0, 4)`
* `IsNumeric(x)` `to Integer.TryParse(x ,out var _)`

Some of these need some checking, and there are obvious other similar ones nearby too that I haven't bothered listing explicitly

GrahamTheCoder:
[...]

PR #283, opened 2019-04-17
Do not convert DateAndTime to DateTime

### LINKED ISSUE #261: VB -> C#: Incorrect class prefix for builtin date functions
### Input code
```vbnet
Public Class Class1
    Sub Foo()
        Dim x = DateAdd("m", 5, Now)
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    public void Foo()
    {
        var x = DateTime.DateAdd("m", 5, DateTime.Now);
    }
}
```

### Expected output
```c#
public class Class1
{
    public void Foo()
    {
        var x = DateAndTime.DateAdd("m", 5, DateTime.Now);
    }
}

```

### Details

Version in use: e.g. 6.6.0

Built in date functions are in the `DateAndTime` class, not `DateTime`.

mrmonday:
Closes #261 

### Problem

When converting VB to C#, built in date functions are prefixed with DateTime rather than DateAndTime.

### Solution

When converting to C#, leave DateAndTime methods alone - these should be converted properly as part of #174.

### DIFF
--- a/ICSharpCode.CodeConverter/Util/ISymbolExtensions.cs
+++ b/ICSharpCode.CodeConverter/Util/ISymbolExtensions.cs
@@ -11,7 +11,10 @@ namespace ICSharpCode.CodeConverter.Util
 #endif
     static class ISymbolExtensions
     {
-        private static readonly string[] TypesToConvertToDateTime = new[] {"DateTime", "DateAndTime" };
+        // A lot of symbols in DateAndTime do not exist in DateTime, eg. DateSerial(),
+        // and some have different names/arguments, eg. DateAdd(). This needs to be handled properly
+        // as part of #174
+        private static readonly string[] TypesToConvertToDateTime = new[] { "DateTime" };
 
         /// <summary>
         /// Checks if 'symbol' is accessible from within 'within'.
--- a/Tests/CSharp/ExpressionTests.cs
+++ b/Tests/CSharp/ExpressionTests.cs
@@ -1203,6 +1203,24 @@ namespace Global.InnerNamespace
         }
 
         [Fact]
+        public void DateTimeToDateAndTime()
+        {
+            TestConversionVisualBasicToCSharpWithoutComments(@"Public Class Class1
+    Sub Foo()
+        Dim x = DateAdd(""m"", 5, Now)
+    End Sub
+End Class", @"using Microsoft.VisualBasic;
+
+public class Class1
+{
+    public void Foo()
+    {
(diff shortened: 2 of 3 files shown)


mrmonday:
It looks like all the line endings changed in `ExpressionTests.cs`. I did a clean clone of the repository, so the endings should have been right... Git always shows `ExpressionsTests.cs` and `NamespaceLevelTests.cs` as modified, regardless of what I did. Not sure what happened there... Sorry I keep getting this wrong...

mrmonday:
I opted to do the easy thing here, and defer cleaning up `Now()` to `DateTime.Now` etc for a later task as part of #174.

GrahamTheCoder:
I think most of your PRs seem to have the right line endings now, and I've got a quickish way to rewrite commits to fix them anyway so don't worry too much.

I think this PR is correct, and I think it's probably the right approach not to try to fix the whole world in a single PR. I'm just hesitating to merge because it does make some cases worse in some sense. I was trying to figure out if there was an easy thing we could do like convert to `DateTime`, but check for the existence of the member before doing so. It's possible to know which members exist in symbolic terms if that helps:

```csharp
private ILookup<string, ISymbol> CreateDateTimeMemberLookup(SemanticModel semanticModel)
{
    return semanticModel.Compilation.GetTypeByMetadataName(typeof(DateTime).FullName).GetMembers().ToLookup(s => s.Name);
}
```

We could do the same for DateAndTime, and create mappings between the symbols to be applied later. That symbol mapping could become the basis for manually adding things in #174.

What do you think?

mrmonday:
I played around with doing that when I worked on this, but I don't really like it as a solution.

 * We need to check the semantics match between the VB and C# methods (they're often subtley different)
 * It only catches a few cases
 * It actually makes the converted code worse when you have date/time heavy code - mixing VB and .NET styles is really ugly compared to just VB or just .NET (at least from what I've seen - obviously this is subjective)

I came to the conclution that a one-to-one conversion is the best thing to do, until a more complete/proper conversion can be done.

GrahamTheCoder:
Yep I think you're right.
For this, let's just make sure converted projects get a reference to Microsoft.visualbasic (I think it might be implicit in vb projects?). Then we can merge it.
In future when we remap, let's do so as an optional pre or post step.

mrmonday:
Opened #299 (via #292) to add the missing reference.

We already output references to `Microsoft.VisualBasic for other things while converting, so I'm not sure it's a blocker for this.

PR #286, opened 2019-04-18
Do not output private get/set for interfaces

### LINKED ISSUE #270: VB -> C#: Incorrect conversion for interface readonly property
### Input code
```vbnet
Public Interface Foo
    ReadOnly Property Something() As String
End Interface
```

### Erroneous output
```c#
public interface Foo
{
    string Something { get; private set; }
}
```

### Expected output
```c#
public interface Foo
{
    string Something { get; }
}
```

### Details
Product in use: VS extension

Version in use: e.g. 6.6.0

mrmonday:
Closes #270 

### Problem

When converting read only or write only properties in interfaces, we should not output the get or set.

### Solution

Special case the output of bodiless get/set for properties in interfaces.

* [x] All tests pass


GrahamTheCoder:
Perfect :)

PR #287, opened 2019-04-18
Convert CDate() correctly

### LINKED ISSUE #268: VB -> C#: CDate() conversion for strings
### Input code
```vbnet
Public Class Class1
    Sub Foo()
        Dim x = CDate("2019-09-04")
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    public void Foo()
    {
        var x = (DateTime)"2019-09-04";
    }
}
```

### Expected output
```c#
public class Class1
{
    public void Foo()
    {
        var x = Convert.ToDateTime("2019-09-04");
    }
}
```

### Details
Product in use: VS extension

Version in use: 6.6.0

I'm not sure how `CDate()` differs compared to `Convert.ToDateTime()` or `DateTime.Parse()` - there may be a better or more accurate thing to replace it with.

mrmonday:
Closes #268

### Problem

When converting `CDate()`, we would previously output a cast, which causes a compile error for `CDate(String)`.

### Solution

VB emits a call to `Microsoft.VisualBasic.CompilerServices.Conversions.ToDate()` - this change does the same.

I do this instead of `Convert.ToDate()`, as there are some differences in implementation which could cause different results.

`Conversions.ToDate()`: https://github.com/Microsoft/referencesource/blob/7f0c570e24e6f0084b3c5794e1a50de9fbe00ca4/Microsoft.VisualBasic/runtime/msvbalib/Helpers/Conversions.vb#L2006-L2079
`Convert.ToDate()`: https://github.com/Microsoft/referencesource/blob/3b1eaf5203992df69de44c783a3eda37d3d4cd10/mscorlib/system/convert.cs#L1712-L1790

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/MethodBodyVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/MethodBodyVisitor.cs
@@ -27,6 +27,7 @@ namespace ICSharpCode.CodeConverter.CSharp
             private readonly SemanticModel _semanticModel;
             private readonly VBasic.VisualBasicSyntaxVisitor<CSharpSyntaxNode> _nodesVisitor;
             private readonly Stack<string> _withBlockTempVariableNames;
+            private readonly HashSet<string> _extraUsingDirectives;
             private readonly HashSet<string> _generatedNames = new HashSet<string>();
 
             public bool IsIterator { get; set; }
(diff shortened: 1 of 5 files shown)


ICSharpCode.CodeConverter/CSharp/MethodBodyVisitor.cs
```diff
@@ -27,6 +27,7 @@ class MethodBodyVisitor : VBasic.VisualBasicSyntaxVisitor<SyntaxList<StatementSy
             private readonly SemanticModel _semanticModel;
             private readonly VBasic.VisualBasicSyntaxVisitor<CSharpSyntaxNode> _nodesVisitor;
             private readonly Stack<string> _withBlockTempVariableNames;
+            private readonly HashSet<string> _extraUsingDirectives;
```
mrmonday: Nice, this is what I wanted to do, but I didn't see an existing way to do it.

GrahamTheCoder: Yeah I was surprised I couldn't find it. Turns out it existed in just the C#->VB direction already.

GrahamTheCoder: Ah, thinking about it, I think I had held off doing this previously due to paranoia about another using statement having conflicting imports. My ideal for this would probably be to add everything fully qualified *and* add the extra using statement. Then at the end do a "simplify names" analysis on the code to shorten names where they are unambiguous. If I could reference someone's existing roslyn fixer for that it'd be pretty easy I think.

Speaking of ambiguity, now I think about it, when creating new variables/methods, really ideally be using `GetUniqueVariableNameInScope` to avoid clashes.

mrmonday: That occurred to me too, completely forgot to go back and fix that though. I'll bear it in mind.

PR #290, opened 2019-04-19
Add parenthesis if necessary for binary expressions

### LINKED ISSUE #262: VB -> C#: Missing parenthesis when converting string concatenation
### Input code
```vbnet
Public Class Class1
    Sub Foo()
        Dim x = "x " & 5 - 4 & " y"
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    public void Foo()
    {
        var x = "x " + 5 - 4 + " y";
    }
}
```

### Expected output
```c#
public class Class1
{
    public void Foo()
    {
        var x = "x " + (5 - 4) + " y";
    }
}
```

### Details

Version in use: e.g. 6.6.0

The converted code gives this error:

```
CS0019	Operator '-' cannot be applied to operands of type 'string' and 'int'
```

Adding parenthesis fixes the error.

mrmonday:
Closes #262 

### Problem

Parenthesis are sometimes missing for binary expressions, causing compilation errors

### Solution

Add parenthesis for binary expressions. Special case some easy examples where they are definitely not required. We would need a full precedence table to work around that.

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -1183,7 +1183,20 @@ namespace ICSharpCode.CodeConverter.CSharp
 
             private static bool PrecedenceCouldChange(VBasic.VisualBasicSyntaxNode node)
             {
-                return node.Parent is VBSyntax.ExpressionSyntax && !(node.Parent is VBSyntax.ArgumentSyntax);
+                bool parentIsSameBinaryKind = node is VBSyntax.BinaryExpressionSyntax && node.Parent is VBSyntax.BinaryExpressionSyntax parent && parent.Kind() == node.Kind();
+                bool parentIsReturn = node.Parent is VBSyntax.ReturnStatementSyntax;
+                bool parentIsLambda = node.Parent is VBSyntax.LambdaExpressionSyntax;
+                bool parentIsNonArgumentExpression = node.Parent is VBSyntax.ExpressionSyntax && !(node.Parent is VBSyntax.ArgumentSyntax);
+                bool parentIsParenthesis = node.Parent is VBSyntax.ParenthesizedExpressionSyntax;
+
+                // Could be a full C# precendence table - this is just a common case
+                bool parentIsAndOr = node.Parent.IsKind(VBasic.SyntaxKind.AndAlsoExpression, VBasic.SyntaxKind.OrElseExpression);
+                bool nodeIsRelationalOrEqual = node.IsKind(VBasic.SyntaxKind.EqualsExpression, VBasic.SyntaxKind.NotEqualsExpression,
+                                                           VBasic.SyntaxKind.LessThanExpression, VBasic.SyntaxKind.LessThanOrEqualExpression,
+                                                           VBasic.SyntaxKind.GreaterThanExpression, VBasic.SyntaxKind.GreaterThanOrEqualExpression);
+                bool csharpPrecedenceSame = parentIsAndOr && nodeIsRelationalOrEqual;
+
+                return parentIsNonArgumentExpression && !parentIsSameBinaryKind && !parentIsReturn && !parentIsLambda && !parentIsParenthesis && !csharpPrecedenceSame;
             }
 
             public override CSharpSyntaxNode VisitLiteralExpression(VBSyntax.LiteralExpressionSyntax node)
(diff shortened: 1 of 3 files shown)


ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -1183,7 +1183,20 @@ private static CSharpSyntaxNode ParenthesizeIfPrecedenceCouldChange(VBasic.Visua
...
             private static bool PrecedenceCouldChange(VBasic.VisualBasicSyntaxNode node)
             {
-                return node.Parent is VBSyntax.ExpressionSyntax && !(node.Parent is VBSyntax.ArgumentSyntax);
+                bool parentIsSameBinaryKind = node is VBSyntax.BinaryExpressionSyntax && node.Parent is VBSyntax.BinaryExpressionSyntax parent && parent.Kind() == node.Kind();
+                bool parentIsReturn = node.Parent is VBSyntax.ReturnStatementSyntax;
+                bool parentIsLambda = node.Parent is VBSyntax.LambdaExpressionSyntax;
+                bool parentIsNonArgumentExpression = node.Parent is VBSyntax.ExpressionSyntax && !(node.Parent is VBSyntax.ArgumentSyntax);
+                bool parentIsParenthesis = node.Parent is VBSyntax.ParenthesizedExpressionSyntax;
+
+                // Could be a full C# precendence table - this is just a common case
```
GrahamTheCoder: Since you're looking at VB syntax to decide precedence I think you'd ideally need both of these:
https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/operator-precedence
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/
One way would be to figure out the old precedence, then the precedence after converting all operators, then parenthesize if different. There's a slightly sneaky possibility though to make this easier. The parser has already told us the VB precedence by the way it parsed the expression. If we convert it, and then try to parse the C#, if we've not parenthesized when we should, then the parse tree will end up a different shape, so when putting together a binary expression, we *could theoretically* reparse what we create and parenthesize if it looks wrong.
That feels like quite a hacky solution, but I wonder if there's a neater way, like existing public helpers that know about precedence of various symbol kinds. Maybe this: https://stackoverflow.com/a/43325150/1128762

GrahamTheCoder: Seems like if we use this [simplifier](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.simplification.simplificationoptions?view=roslyn-dotnet) it might also work for things like cutting down fully qualified names. Worth a go I'd say!

GrahamTheCoder: Rather than actually try to handle adding parenthesis ourselves everywhere, and adding fully qualified names everywhere, we may be able to use the Simplifier's [expand](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.simplification.simplifier.expand?view=roslyn-dotnet#Microsoft_CodeAnalysis_Simplification_Simplifier_Expand_Microsoft_CodeAnalysis_SyntaxToken_Microsoft_CodeAnalysis_SemanticModel_Microsoft_CodeAnalysis_Workspace_System_Func_Microsoft_CodeAnalysis_SyntaxNode_System_Boolean__System_Threading_CancellationToken_) capability on the input. So long as we can get a semantic model out of it again we could do the entire conversion from that, then call reduce at the end.

mrmonday: Yes, I was looking at both when I was working on it. I decided it needed too much thought and decided to handle the easy/common cases 😛 

Always outputting parenthesis and simplifying afterwards is probably a nice approach, I'll take a look at it next week.

I'm a little more reluctant on the FQN front - I'm worried about how it will turn out when there's less semantic information available. It might be fine, but the test suite definitely seems weaker in the "missing relevant code" conversions. Definitely something to look into though.

mrmonday: I've been looking into using `Simplifier`, I'm struggling to get anywhere though.

From what I can gather, we need to add the relevant annotations (no problem), then call one of the [`ReduceAsync()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.simplification.simplifier.reduceasync?view=roslyn-dotnet#Microsoft_CodeAnalysis_Simplification_Simplifier_ReduceAsync_Microsoft_CodeAnalysis_Document_Microsoft_CodeAnalysis_Options_OptionSet_System_Threading_CancellationToken_) methods. These all require a `Document`, which we don't currently seem to have anywhere - it looks like we just format then output the resulting syntax tree.

I'm not familiar enough with Roslyn's machinery to figure out how to put this together.

GrahamTheCoder: It's possible to create a containing project and document on the AdhocWorkspace. Though if you use `Reduce` instead of `ReduceAsync` it isn't needed. The tests totally fail because I think I've broken the connection between the syntax tree and semantic model somewhere along the way. Have a look and see if it's enough to help you make progress. If not I'll do some more experimentation when I have time.

Once we've figured it out, we'll probably find the performance sucks. The trick there I think will be to minimize the number of new compilations we create (since all the documents are independent we should be able to just recreate it a single time hopefully).

PR #299, opened 2019-04-22
Add a reference to Microsoft.VisualBasic for VB -> C#

### LINKED ISSUE #292: VB -> C#: Converter does not add project reference to Microsoft.VisualBasic
When converting, some converted code references `Microsoft.VisualBasic`. If this happens, there should be a reference to this DLL added to the `.csproj` that was output.

mrmonday:
Closes #292

### Problem

Converting from VB to C# uses things from the `Microsoft.VisualBasic` namespace. We need to add a reference to the assembly so this does not error.

### Solution

Adds an extra bit of regex to add in the reference if it is not already there. Doesn't seem like the nicest solution, but it seems like regex is used for the rest of the `.csproj` conversion, so this is probably fine.

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/VBToCSConversion.cs
+++ b/ICSharpCode.CodeConverter/CSharp/VBToCSConversion.cs
@@ -11,6 +11,7 @@ using SyntaxFactory = Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
 using CSSyntax = Microsoft.CodeAnalysis.CSharp.Syntax;
 using SyntaxKind = Microsoft.CodeAnalysis.VisualBasic.SyntaxKind;
 using VBSyntax = Microsoft.CodeAnalysis.VisualBasic.Syntax;
+using System.Text.RegularExpressions;
 
 namespace ICSharpCode.CodeConverter.CSharp
 {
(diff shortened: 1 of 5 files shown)


mrmonday:
I looked into adding the reference to the compilation in `ProjectConversion.cs`  as you mentioned, but I wasn't sure how to test it properly and my solution seemed rather hacky. I used something like:

```c#
var path = Path.Combine(Path.GetDirectoryName(typeof(object).GetAssemblyLocation()), "Microsoft.VisualBasic.dll");
var reference = MetadataReference.CreateFromFile(path);
```

This didn't match the path of the other references which might cause issues - I wasn't sure of the best way to find the right path for the target framework.

In my quick tests it seemed like the `Microsoft.VisualBasic` reference was already available in the compilation - I don't know if that was specific to the test I was running, or whether it is always there (in which case there's nothing that needs to be done).

GrahamTheCoder:
Yep you're right - it looks like the implicit reference to visualbasic in the vbproj becomes just a normal reference, so it already gets copied across to the convertedCompilation 🎉 

If you come across a sensible API for the project files then do let me know, but until then, regexes are fine for the number of project file adjustments we currently have.

Issue #295, opened 2019-04-19
VB -> C#: Assignment return support for properties


mrmonday:
### Input code
```vbnet
Public Class Class1
    Public ReadOnly Property Foo() As String
        Get
            Foo = ""
        End Get
    End Property
End Class
```

### Erroneous output
```c#
public class Class1
{
    public string Foo
    {
        get
        {
            Foo = "";
        }
    }
}
```

### Expected output
```c#
public class Class1
{
    public string Foo
    {
        get
        {
            string FooRet;
            FooRet = "";
            return FooRet;
        }
    }
}
```

GrahamTheCoder:
That is rather surprising behaviour. If there's a setter too, would it use that to actually assign to Foo and then return default? Or still do the same thing?
Sounds like a good one for a self verifying test in ExpressionTests.vb so someone like me doesn't come along with the wrong idea about what happens and try to "fix" it.

mrmonday:
https://github.com/icsharpcode/CodeConverter/pull/302/files#r278278050

Code in case it gets lost:

```vbnet
Imports System
				
Public Module Module1
	Public Sub Main()
		Console.WriteLine(X) '8
	End Sub
	
	Public ReadOnly Property X As String
		Get
			X = 4
			X = X * 2
			Dim y = "random variable to check it isn't just using the value of the last statement"
		End Get
	End Property
	
End Module
```

GrahamTheCoder:
Just need to add the test, then change from MethodBlockSyntax to MethodBlockBaseSyntax here:
https://github.com/icsharpcode/CodeConverter/pull/302#discussion_r278773985
If it works then yay, merge, otherwise debug.

PR #302, opened 2019-04-24
Handle more cases with VB implicit return

### LINKED ISSUE #293: VB -> C#: Missed cases for assignment return
Missed some cases for the new VB assignment return conversions.

### Input code
```vbnet
Public Class Class1
    Protected Overrides Function Foo() As String
        Foo = Foo & ""
    End Function
End Class
```

### Erroneous output
```c#
public class Class1
{
    protected override string Foo()
    {
        Foo = Foo + "";
        return default(string);
    }
}
```

### Expected output
```c#
public class Class1
{
    protected override string Foo()
    {
        string FooRet;
        FooRet = FooRet + "";
        return FooRet;
    }
}
```

mrmonday:
Closes #293

### Problem

Some cases were missed when handling VB's assignment return syntax.

### Solution

* Ensure we look at the current node in the assignment case.
* When we have an identifier on the right hand side of an expression, replace it with the return variable, unless it is part of an invocation.

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -669,7 +669,7 @@ namespace ICSharpCode.CodeConverter.CSharp
                 if (!AllowsImplicitReturn(node)) return null;
 
                 string methodName = node.SubOrFunctionStatement.Identifier.ValueText;
-                bool assignsToMethodNameVariable = node.Statements.Any(s => s.DescendantNodes()
+                bool assignsToMethodNameVariable = node.Statements.Any(s => s.DescendantNodesAndSelf()
                     .OfType<VBSyntax.AssignmentStatementSyntax>().Any(assignment =>
                         (assignment.Left as VBSyntax.SimpleNameSyntax)?.Identifier.ValueText.Equals(methodName,
                             StringComparison.OrdinalIgnoreCase) == true));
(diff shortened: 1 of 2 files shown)


ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -1855,18 +1855,31 @@ public override CSharpSyntaxNode VisitIdentifierName(VBSyntax.IdentifierNameSynt
                                                     || node.Parent is VBSyntax.MemberAccessExpressionSyntax maes && maes.Expression == node
                                                     || node.Parent is VBSyntax.QualifiedNameSyntax qns && qns.Left == node
                     ? QualifyNode(node, identifier) : identifier;
-                return AddEmptyArgumentListIfImplicit(node, qualifiedIdentifier);
+
+                var withArgList = AddEmptyArgumentListIfImplicit(node, qualifiedIdentifier, out bool addedArgumentList);
+                if (!addedArgumentList && !node.Parent.IsKind(VBasic.SyntaxKind.InvocationExpression)) {
+                    var vbMethodBlock = node.Ancestors().OfType<VBSyntax.MethodBlockSyntax>().FirstOrDefault();
```
GrahamTheCoder: I think this needs to be `MethodBlockBaseSyntax` to account for this case:
https://dotnetfiddle.net/FpGqfE

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -1855,18 +1855,31 @@ public override CSharpSyntaxNode VisitIdentifierName(VBSyntax.IdentifierNameSynt
                                                     || node.Parent is VBSyntax.MemberAccessExpressionSyntax maes && maes.Expression == node
                                                     || node.Parent is VBSyntax.QualifiedNameSyntax qns && qns.Left == node
                     ? QualifyNode(node, identifier) : identifier;
-                return AddEmptyArgumentListIfImplicit(node, qualifiedIdentifier);
+
+                var withArgList = AddEmptyArgumentListIfImplicit(node, qualifiedIdentifier, out bool addedArgumentList);
+                if (!addedArgumentList && !node.Parent.IsKind(VBasic.SyntaxKind.InvocationExpression)) {
```
GrahamTheCoder: Probably need to watch out for `AddressOf Foo` as well.
More nebulous: Technically `NameOf(Foo)` probably shouldn't change in case someone depends on it.
I wonder if we can use the symbol type instead to check for anything with that name that isn't equal to the method's symbol? It might avoid the need for the special case below too where you need to know if it was turned into an invocation.

mrmonday: I've updated this to use the symbol information instead - this is far nicer.

I've still special-cased `NameOf` - it's semantically different, but produces the same results as in VB. I don't know if this is desired or not.

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
GrahamTheCoder: The symbol thing looks much better, I think you still need to switch to `OfType<MethodBlockBaseSyntax>`  here though.

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
```diff
@@ -669,7 +669,7 @@ private IdentifierNameSyntax GetRetVariableNameOrNull(VBSyntax.MethodBlockSyntax
                 if (!AllowsImplicitReturn(node)) return null;
 
                 string methodName = node.SubOrFunctionStatement.Identifier.ValueText;
-                bool assignsToMethodNameVariable = node.Statements.Any(s => s.DescendantNodes()
+                bool assignsToMethodNameVariable = node.Statements.Any(s => s.DescendantNodesAndSelf()
```
GrahamTheCoder: Probably worth a quick check to see if we can use the data flow analysis for `ReadInside` and `WrittenInside` or just `VariablesDeclared` to check for the method name in there. If not though I'll happily merge it like this.

mrmonday: I've updated the code to use data flow analysis instead for this.

ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
GrahamTheCoder: Ah I see, this is a separate issue and we can address it separately (https://github.com/icsharpcode/CodeConverter/issues/295)

PR #303, opened 2019-04-25
Add more missing initialisations for VB to C#

### LINKED ISSUE #297: VB -> C#: Variable not initialised with multiple declarations
### Input code
```vbnet
Public Class Class1
    Public Sub Foo()
        Dim i As Integer, temp As String = String.Empty
        i += 1
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    public void Foo()
    {
        int i;
        string temp = string.Empty;
        i += 1;
    }
}
```

### Expected output
```c#
public class Class1
{
    public void Foo()
    {
        int i = 0;
        string temp = string.Empty;
        i += 1;
    }
}
```

### Details

This seems like a missed case from #285 / #263.

mrmonday:
Closes #297 

### Problem

Variables are not initialised when convertion from C# to VB when things other than simple assignments are used.

### Solution

The AlwaysAssigned dataflow analysis doesn't handle `+=`. Explicitly check for simple assignments when deciding to leave a variable uninitialised.

Unfortunately it doesn't seem like we can handle the out variable case, as the symbol info doesn't include `out` for VB - so there's a slight regression in output there.

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
+++ b/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
@@ -52,17 +52,10 @@ namespace ICSharpCode.CodeConverter.CSharp
                 EqualsValueClauseSyntax equalsValueClauseSyntax;
                 if (adjustedInitializer != null) {
                     equalsValueClauseSyntax = SyntaxFactory.EqualsValueClause(adjustedInitializer);
-                } else if (isField) {
+                } else if (isField || !PossiblyWrittenBeforeRead(declarator, name)) {
                     equalsValueClauseSyntax = null;
                 } else {
-                    bool readBeforeWritten = CouldBeWrittenBeforeRead(declarator, name);
-
-                    // 3. We default to outputting an initializer. We only skip this if we can show the variable is never read before it is written
-                    if (!readBeforeWritten) {
-                        equalsValueClauseSyntax = null;
-                    } else {
-                        equalsValueClauseSyntax = SyntaxFactory.EqualsValueClause(SyntaxFactory.DefaultExpression(type));
-                    }
+                    equalsValueClauseSyntax = SyntaxFactory.EqualsValueClause(SyntaxFactory.DefaultExpression(type));
                 }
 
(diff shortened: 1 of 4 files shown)


GrahamTheCoder:
I think you might have got a bit mixed up about what the data flow analysis is doing (apologies if it's in fact me who's confused). I've pushed a couple of tests which highlight what I think is going on (see commit comment with them).

In the CSharp compiler you can see that tracking definite assignment is a bit involved:
http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/FlowAnalysis/DefiniteAssignment.cs,d5ba0ba744faa675,references

However, we only need a small set of that behaviour to satisfy this use case. I didn't read the class, but I suspect if you did, you'd find it's keeping track of where each variable is first potentially/definitely read and potentially/definitely written (on each possible code path through the method).
Since we're only doing any of this analysis to neaten things up when it's not used, we should err on the side of outputting the initialization. Here's a PR that relates to the code fix we're essentially shortcutting on the result (Value assigned to 'i' is never used, IDE0059): https://github.com/dotnet/roslyn/pull/34074/files

I wonder if we can use inspiration from, or directly use anything of that. Similarly to the way that usings are tidied in the second pass, perhaps this diagnostic can be detected and acted upon? If we only want to do so where we've added it, we could annotate the node at the time of adding and check for that later when we have a C# semantic model handy.

Sorry I don't have more time to dig into this, possibly for a few days.

mrmonday:
You're completely right, I misunderstood how the data flow analysis worked.

I took a look at always emitting defaults, then cleaning up after, I couldn't figure out how to get Roslyn to produce IDE diagnostics alongside the other errors.

I have something using data flow analysis working locally now, I want to clean it up some more though before I push it up. Thanks for your help!

mrmonday:
I've cleaned up this logic quite a lot, hopefully it's good now.

GrahamTheCoder:
So much neater - looks great. I've created SemanticModelExtensions to house this kind of neat isolated analysis. Thanks!

PR #306, opened 2019-05-01
Enum implicit cast support

### LINKED ISSUE #180: VB -> C# Enum implicit cast
### Input code
```vbnet
Public Enum TestEnum As Integer
    Test1 = 0
    Test2 = 1
End Enum

Dim t1 As Integer = TestEnum.Test1
```
### Erroneous output
```C#
int t1 = TestEnum.Test1;
```
### Expected output
```C#
int t1 = (int)TestEnum.Test1;
```
### Details
Product in use: VS Extension
Version in use: 6.0.0

In VB.Net, Enums have an implicit cast to their corresponding value type.

mrmonday:
Closes #180

### Problem

We currently don't handle implicit enum conversions. This adds support for them

### Solution

 * Add support for explicit conversions in variable declarations. This is a little verbose for eg. floating point equals integer literal.
 * Add support for enum -> integer conversions. This is a little verbose for eg. Binary equals. We could special case this in `VisitBinaryExpression()` to give nicer output, I opted not to do that here - happy to change it though.
 * Not sure about how I got the `TypeConversionAnalyzer` into `CommonConversions` - I tried a few ways and they all seemed kinda ugly.

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
+++ b/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
@@ -26,6 +26,32 @@ using CSSyntax = Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace ICSharpCode.CodeConverter.CSharp
 {
+    internal class VbSyntaxNodeExtensions
+    {
+        public static ExpressionSyntax ParenthesizeIfPrecedenceCouldChange(Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxNode node, ExpressionSyntax expression)
+        {
+            return PrecedenceCouldChange(node) ? SyntaxFactory.ParenthesizedExpression(expression) : expression;
+        }
+
+        public static bool PrecedenceCouldChange(Microsoft.CodeAnalysis.VisualBasic.VisualBasicSyntaxNode node)
+        {
+            bool parentIsSameBinaryKind = node is VBSyntax.BinaryExpressionSyntax && node.Parent is VBSyntax.BinaryExpressionSyntax parent && parent.Kind() == node.Kind();
+            bool parentIsReturn = node.Parent is VBSyntax.ReturnStatementSyntax;
+            bool parentIsLambda = node.Parent is VBSyntax.LambdaExpressionSyntax;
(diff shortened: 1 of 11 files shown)


ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
```diff
@@ -51,7 +52,9 @@ public CommonConversions(SemanticModel semanticModel, VisualBasicSyntaxVisitor<C
                 bool isField = declarator.Parent.IsKind(SyntaxKind.FieldDeclaration);
                 EqualsValueClauseSyntax equalsValueClauseSyntax;
                 if (adjustedInitializer != null) {
-                    equalsValueClauseSyntax = SyntaxFactory.EqualsValueClause(adjustedInitializer);
+                    var vbInitializer = declarator.Initializer?.Value;
+                    var convertedInitializer = vbInitializer == null ? adjustedInitializer : TypeConversionAnalyzer.AddExplicitConversion(vbInitializer, adjustedInitializer);
```
mrmonday: Explicit conversions are never needed for `AsClause`, since the type is inferred from the RHS. Probably should have left a comment about that here.

Tests/CSharp/StatementTests.cs
```diff
@@ -1403,7 +1403,7 @@ public bool CanDoWork(object Something)
     {
         switch (true)
         {
-            case object _ when (DateAndTime.Today.DayOfWeek == DayOfWeek.Saturday) | (DateAndTime.Today.DayOfWeek == DayOfWeek.Sunday):
+            case object _ when ((int)DateAndTime.Today.DayOfWeek == (int)DayOfWeek.Saturday) | ((int)DateAndTime.Today.DayOfWeek == (int)DayOfWeek.Sunday):
```
GrahamTheCoder: These are both enums of the same type so I'm surprised converting them was necessary

mrmonday: The difficulty is we handle enum<->int in the type conversion analysis, rather than in the binary expression visitor. `typeInfo.Type` is the enum type, and `typeInfo.ConvertedType` is int, so we insert the cast (which is required everywhere else). This makes sense, since presumably VB just does an integer comparison for enums. Fine for the compiler, kind of ugly for converted code.

There's a few ways we could handle this, I haven't put too much thought into it though.

Issue #281, opened 2019-04-16
VB -> C#: Field initializers referencing non-static fields


mrmonday:
### Input code
```vbnet
Public Class A
    Private x As Integer = 2
    Private y(x) As Integer
End Class
```

### Erroneous output
```c#
public class A
{
    private int x = 2;
    private int[] y = new int[x + 1];
}
```

### Expected output
```c#
public class A
{
    private int x = 2;
    private int[] y;
    public A()
    {
        y = new int[x + 1];
    }
}
```

### Details
Product in use: VS extension

Version in use: 6.6.0

Accessing a non-static field in C# is an error, where it allowed in VB ([CS0236](https://docs.microsoft.com/en-us/dotnet/csharp/misc/CS0236)).

It's a bit more complicated than this, since you need to account for multiple constructors.

GrahamTheCoder:
Minimum: Add the same initialization to all bottom-level constructors (i.e. which don't pass on to another of the type's constructors). 
Nice to have: Try to create a new method or constructor to add to bottom level constructors in order to avoid a bit of duplication.

There's another related issue that I haven't looked in to. Which is that I think VB has a different initialization order to C# in terms of constructors and fields. There's [this handy article](https://blogs.msdn.microsoft.com/ericlippert/2008/02/15/why-do-initializers-run-in-the-opposite-order-as-constructors-part-one/) that explains what's going on with C# and why. I don't know what the VB order is, and whether its differences affect this. I can't immediately see a comparison online.

tverweij:
@GrahamTheCoder 
" I don't know what the VB order is"

The VB Initializers run _within the constructor_ before _any other constructor code_ in the sequence top to bottom as they are defined in the class.

GrahamTheCoder:
Thanks, that sounds quite straightforward to implement. So when there's a base class it's presumably:
1. Base fields (in declaration order)
1. Base constructor
1. This fields (in declaration order)
1. This constructor

tverweij:
Exactly

GrahamTheCoder:
OK so in C# it's:
1. Base fields (in declaration order)
1. This fields (in declaration order)
1. Base constructor
1. This constructor

This gives the possibility that in VB a field in "this" tries to use something modified in the base constructor.
The easiest way to make that work in C# would be for any class that inherits from another, move all its initializers into the constructor in the C# code to emulate what VB did.

[...]

PR #307, opened 2019-05-01
Support more complex ByRef arguments

mrmonday:
References #264

Add support for more complicated arguments to methods with ByRef parameters.

Takes a different approach to #204.

### Problem

VB supports passing anything as a ByRef parameter, where C# requires something it can take the address of.

### Solution

 * When generating arguments, if the parameter is `ByRef`, create an identifier with a random name, annotate it (with `AdditionalLocal.Annotation`), and add an `AdditionalLocal`
 * When visiting an expression statement, generate a sensible unique variable name, insert additional local declarations, and update the name of the argument.
 * When calling a method with `ByRef` parameters in a field initializer, generate a method without `ByRef` to call instead
 * We only handle the easiest of cases for field initializers - a lot of the functionality needs to overlap with #281. We throw errors for some (not all) of the more complex cases. The rest should fail to compile in the generated C# anyway.
 * Add some additional type conversion support, for explicit upcasting. VB supports `ByRef` for mismatching types - we need to insert an explicit cast (and therefore, an additional local variable).

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/TypeConversionAnalyzer.cs
+++ b/ICSharpCode.CodeConverter/CSharp/TypeConversionAnalyzer.cs
@@ -24,9 +24,9 @@ namespace ICSharpCode.CodeConverter.CSharp
 
         private CommonConversions CommonConversions { get; }
 
-        public ExpressionSyntax AddExplicitConversion(Microsoft.CodeAnalysis.VisualBasic.Syntax.ExpressionSyntax vbNode, ExpressionSyntax csNode, bool addParenthesisIfNeeded = false)
+        public ExpressionSyntax AddExplicitConversion(Microsoft.CodeAnalysis.VisualBasic.Syntax.ExpressionSyntax vbNode, ExpressionSyntax csNode, bool addParenthesisIfNeeded = false, bool alwaysExplicit = false)
         {
-            var conversionKind = AnalyzeConversion(vbNode, csNode, out var vbConvertedType);
+            var conversionKind = AnalyzeConversion(vbNode, csNode, alwaysExplicit, out var vbConvertedType);
             csNode = addParenthesisIfNeeded && conversionKind == TypeConversionKind.Implicit
                 ? CommonConversions.ParenthesizeIfPrecedenceCouldChange(vbNode, csNode)
                 : csNode;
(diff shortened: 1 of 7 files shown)


mrmonday:
Just tried this on a large codebase, looks like I missed a couple of easy cases - I'll take a look.

mrmonday:
Changing this to WIP again - there are far more places that need to be updated to handle this properly. I think it's best to introduce another visitor for this, will play around with it and see how it goes.

GrahamTheCoder:
Just in case it's relevant, one idea I was considering, is to return more information from existing visitors.

Another vague plan is to maintain a mapping of nodes for use so that in the existing second pass, the vb node and semantic model would be available alongside the first pass C# node and semantic model.

mrmonday:
I'm pretty happy with this now.

There's one additional case which this doesn't cover, but I'd rather open a separate issue for it (like handling fields properly).

Something like:

```vbnet
If False AndAlso MethodWithByRef(MethodWithSideEffects()) Then
    DoSomething()
End If
```

Will become:
```c#
var arg = MethodWithSideEffects();
if (false && MethodWithByRef(ref arg))
{
    DoSomething();
}
```

Causing the side-effects to happen even when they shouldn't. Dealing with that is a bit involved, and this PR is already quite large.

ICSharpCode.CodeConverter/CSharp/TypeConversionAnalyzer.cs
```diff
@@ -68,6 +68,9 @@ private TypeConversionKind AnalyzeConversion(Microsoft.CodeAnalysis.VisualBasic.
 
             if (csType is null || csConvertedType is null)
             {
+                if (alwaysExplicit && vbType != vbConvertedType) {
```
GrahamTheCoder: We only return implicit if "alwaysExplicit" is set...sounds slightly backwards to me. I expect it's a naming thing?

GrahamTheCoder:
Looks good. Yes the issue you mention would definitely be good to fix. especially for the case where the first condition is a null check and the second condition uses that variable which is quite a common pattern. I agree that can be a separate PR though.

ICSharpCode.CodeConverter/CSharp/TypeConversionAnalyzer.cs
mrmonday: I meant it as "always insert an explicit type conversion". I agree it's a bit confusing.

mrmonday:
Opened #310 for the conditional case, and added a comment to #281 for the field case.

PR #309, opened 2019-05-03
Implement correct string comparisons for VB

### LINKED ISSUE #308: VB -> C#: String comparison behaviour does not match VB
### Input code
```vbnet
Class Class1
    Sub Main()
        Dim s1 As String = Nothing
        Dim s2 As String = ""
        If s1 <> s2 Then
            Throw New Exception()
        End If
    End Sub
End Class
```

### Erroneous output
```c#
public class Class1
{
    public static void Main()
    {
        string s1 = null;
        string s2 = "";
        if (s1 != s2)
            throw new Exception();
    }
}
```

### Expected output
```c#
using Microsoft.VisualBasic.CompilerServices;

public class Class1
{
    public static void Main()
    {
        string s1 = null;
        string s2 = "";
        if (Operators.CompareString(s1, s2, TextCompare: false) != 0)
            throw new Exception();
    }
}
```

### Details

In VB, the code does not throw. In the current C#, it does throw.

In VB, `null` and `string.Empty` compare as the same.

The value of the `TextCompare` parameter depends on whether `Option Compare Text` or `Option Compare Binary` is set (I believe it defaults to Binary).

Unless we're dealing with literals, I don't think there's any way to automatically clean this up.

Source for `CompareString`: https://github.com/dotnet/corefx/blob/90dc9459b8eaf0c8ac93440e58c7535c67dacc22/src/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/Operators.vb#L845-L883

mrmonday:
Closes #308

### Problem

String comparison behaviour for converted VB -> C# code does not match VB's behaviour.

### Solution

Added a special case for string equality. Added a special case to that to produce `string.IsNullOrEmpty()` when we can to produce slightly nicer code.

* [x] All tests pass

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs
@@ -37,6 +37,7 @@ namespace ICSharpCode.CodeConverter.CSharp
             private static readonly Type ExtensionAttributeType = typeof(ExtensionAttribute);
             private readonly TypeConversionAnalyzer _typeConversionAnalyzer;
             public CommentConvertingNodesVisitor TriviaConvertingVisitor { get; }
+            private bool _optionCompareText = false;
 
             private CommonConversions CommonConversions { get; }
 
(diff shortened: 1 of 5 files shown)


mrmonday:
Weird, the tests passed locally... Guess it didn't actually run them all. I'll fix them on Monday, unless you get to it before me.

GrahamTheCoder:
Also see https://github.com/icsharpcode/CodeConverter/compare/string-comparison?expand=1

mrmonday:
I've rebased and pushed fixed the test case.

I think your branch handles more cases of implicit conversions, which this doesn't.

Unfortunately I don't think we can do much more than this PR does, in terms of not calling the method - in general it's impossible to know if a variable will be null at runtime (at least until we have non-nullable types). Without that knowledge, we have to call the method (or effectively duplicate the logic). I can think of two additional cases where we could handle this differently:

 * Comparison against any string or character literal could become `==`/`!=`
 * Comparison against "known null or empty" values, beyond just literals (though this produces potentially unintuitive output)

I'll add some code to handle the first case.

GrahamTheCoder:
Thanks!

PR #346, opened 2019-08-04
Avoid unnecessary parentheses/qualification etc.

### LINKED ISSUE #344: Use ReduceAsync to simplify output code
This will allow a simplification of code in the converter - which will be able to just always output the most verbose, fully qualified, over-parenthesized version of everything, then have it simplified at the end. This may also make TidyUsings and the SecondPass redundant.

I've spiked the vb->cs case when a project object is available here:
https://github.com/icsharpcode/CodeConverter/compare/ReduceAsync?expand=1

The next thing is to create a workspace and project from a purely textual input in the ConvertText method.

Finally we can enforce passing a document to ConvertSingle, and thus avoid the need to pass ContainingProject since the document references the project. This would cause source-level breaks for public api users, so it may be worth creating deprecated shim methods with the old signature for backwards compatibility.

GrahamTheCoder:
Fixes #344

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/CompilationErrorFixer.cs
+++ b/ICSharpCode.CodeConverter/CSharp/CompilationErrorFixer.cs
@@ -24,72 +24,30 @@ namespace ICSharpCode.CodeConverter.CSharp
         public CSharpSyntaxNode Fix()
         {
             var syntaxNode = _syntaxTree.GetRoot();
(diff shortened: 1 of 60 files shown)


GrahamTheCoder:
Generating the extra compilations required for these improvements does cause a performance degradation. I'll come back to it if it seems problematic. Following these major changes, some full system testing should happen before next release (i.e. Try all the menu items and the web app on a few interesting cases)

PR #367, opened 2019-08-31
Enum and initializer improvements

GrahamTheCoder:
Covers a multitude of issues around enums and intializers. They're only really related because the fix for some of them compounds the effects of others if not also fixed. There are a few more semi-related issues I want to cover here still:
* [x] Test+fix lambdas getting assigned with var - they need an explicit type (or constructor) in C#
  * The simplifier null refs in one test implying the syntax tree is invalid for the lambda body
  * Need to ensure co-ordination between two functions is done only when needed (not actions/funcs), is robust and well documented - see TODOs in CreateLambdaExpression
* [ ] Test+fix issue with no conversion available for e.g. `MyEnum x = MyEnum.Parse(...)` (check VB IL)
* [ ] Test+fix issue with overqualifying types on fields/locals (compounded in this PR - add simplifier annotation in more places?)

I've raised https://github.com/dotnet/roslyn/issues/38439 to track the request to introduce an API for the property I've grabbed via reflection.


GrahamTheCoder:
In response to: https://github.com/dotnet/roslyn/issues/38439#issuecomment-526919600
> I would argue that
> 
> ```vbnet
> Dim x as Integer
> ```
> 
> should be translated to
> 
> ```cs
> int x = default(int);
> ```
> 
> As per Visual Basic specification stating "All locals are initialized to their type's default value.", see https://docs.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/statements#local-variables-and-parameters.

### Specific case of initializers
As far as I know VB and C# don't do anything to default initialize variables in the sense no extra IL instructions are generated to do that, variables just start initialized to all zeroes. VB and C# make different choices on whether/how to help users avoid bugs around initialization at compile time.

Consider the case:
```vbnet
Public Function Check() As Boolean
    Dim valType as Integer
    Dim refType as Object
    Dim refType2 as Object
    valType = 3
    refType = refType2 ' Warning BC42104 Variable 'refType2' is used before it has been assigned a value. A null reference exception could result at runtime.
    Return Equals(valType, refType)
End Function
```

My assumption of what you propose as the conversion:
```csharp
public bool Check()
{
    int valType = default(int); //Message IDE0059 Unnecessary assignment of a value to 'valType'
    object refType = default(object);  //Message IDE0059 Unnecessary assignment of a value to 'refType'
    object refType2 = default(object);
    valType = 3;
    refType = refType2;
    return Equals(valType, refType);
}
```

The CSharp conversion contains redundant code that causes the compiler to issue "Messages". That doesn't seem like a faithful representation of the VB which contained no redundant code or "Messages".

My proposed conversion:
```csharp
public bool Check()
{
    int valType;
    object refType;
    object refType2 = default(object);
    valType = 3;
    refType = refType2;
    return Equals(valType, refType);
}
```
It's a bit weird in both solutions that the converter "fixes" a compiler warning. Basically to me warnings/messages indicate non-idiomatic code, so if the idiomatic code can be determined at conversion time, and functions identically at runtime, I'd prefer it.

### General priorities

There are several competing priorities for the converter (I'll add these to the wiki or something later):
The output code should aim to:
1. Function the same as input code at runtime
1. Present the same public API
1. Be idiomatic for C# (i.e. If a C# developer wrote the same code by hand, how would they write it?)
1. Be easy to refactor to even more idiomatic C# (at the cost of changing runtime behaviour)
1. Be a recognisable conversion of the input code

I'm treating them in roughly that priority order, though that's open to discussion and change. Thanks for your thoughts!

PR #381, opened 2019-09-22
Set up CI with Azure Pipelines

GrahamTheCoder:


christophwille:
Maybe take a look at https://github.com/icsharpcode/ILSpy/blob/master/azure-pipelines.yml regarding vmImage. Also most likely this is also a netcore build instead of desktop.

Issue #373, opened 2019-09-06
VB -> C#: Error with multiple variable declarations with implicit type


mrmonday:
### Input code
```vbnet
Friend Module Module1
    Sub Main()
        Dim x, y As Date
        Console.WriteLine(x)
        Console.WriteLine(y)
    End Sub
End Module
```

### Erroneous output
```c#
using System;

namespace ConsoleApp3
{
    internal static class Module1
    {
        public static void Main()
        {
            var x = default(DateTime), y = default(DateTime);
            Console.WriteLine(x);
            Console.WriteLine(y);
        }
    }
}
```

### Expected output
```c#
using System;

namespace ConsoleApp3
{
    internal static class Module1
    {
        public static void Main()
        {
            DateTime x = default(DateTime), y = default(DateTime);
            Console.WriteLine(x);
            Console.WriteLine(y);
        }
    }
}
```

Alternatively, this could be split into two declarations.

### Details

Error:

```
error CS0819: Implicitly-typed variables cannot have multiple declarators
```

Version used: master @ 3ffc597ffdce991ef0cccf01bd7c884b9d337d6a

BrianFreemanAtlanta:
In looking at the documentation here:
https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/variables/variable-declaration
I believe that in VB the statement 
```
Dim x, y as Date
```
both x and y  are strongly typed as Date variable type so the expected output would be
```
using System;

namespace ConsoleApp3
{
    internal static class Module1
    {
        public static void Main()
        {
            DateTime x = default(DateTime);
            DateTime y = default(DateTime);
            Console.WriteLine(x);
            Console.WriteLine(y);
        }
    }
}
```
Because the intent of both x and y being typed to DateTime and not var are then being met.  In addition when testing the first option with "," separator in C# is not valid syntax so a 2 line declaration is needed.

So I believe for this issue the expected output should be revised.

GrahamTheCoder:
Thanks for your interest. You're correct that both x and y should be DateTimes. Both the original, and your suggestion achieve that, and are acceptable conversions. I'd happily accept a PR for either, with a marginal preference for the one mentioned in the current issue description.
This is simply because it looks closer to the VB and is equal in all other respects I can think of such as these [key points](https://github.com/icsharpcode/CodeConverter/blob/master/.github/CONTRIBUTING.md#deciding-what-the-output-should-be).

[...]

PR #385, opened 2019-10-04
Change for issue #373. Multiple declaration with implicit type.

BrianFreemanAtlanta:
Link to issue(s) this covers
closes issue #373 .
### Problem
Current conversion uses var for declarations with multiple variables but C# issues message that multiple declaration requires an explicit type.

### Solution
I made a change in SplitVariableDeclarations to requireExplicitType for the first declaration when multiple declarations are involved.

Added test `DeclarationStatementTwoVariables` to confirm correction.  Modified two affected tests after confirming they conform to the same pattern.  Changed `ForWithBlock` and `ForWithSingleStatement`
* [X] All tests pass


GrahamTheCoder:
Looks good to me! I've just moved the logic to be calculated before the loop and thus not depend on span start. I couldn't think of a case where we'd want it to be different for each declarator, and no tests failed by doing so. If there was a particular case you were thinking of, do let me know.

Thanks again for your contribution. Have a look at the "good first issue"s if you're interested in doing more:
https://github.com/icsharpcode/CodeConverter/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

Issue #365, opened 2019-08-21
VB -> C#: Missing conversion for array index assignment


mrmonday:
### Input code
```vbnet
Module Module1
    Sub Main()
        Dim x As New ArrayList
        x.Add("a")

        Dim xs(1) As String

        xs(0) = x(0)

    End Sub
End Module
```

### Erroneous output
```c#
namespace ConsoleApp10
{
    static class Module1
    {
        public static void Main()
        {
            ArrayList x = new ArrayList();
            x.Add("a");

            string[] xs = new string[2];

            xs[0] = x[0];
        }
    }
}
```

### Expected output
```c#
namespace ConsoleApp10
{
    static class Module1
    {
        public static void Main()
        {
            ArrayList x = new ArrayList();
            x.Add("a");

            string[] xs = new string[2];

            xs[0] = Conversions.ToString(x[0]);
        }
    }
}
```

### Details

Version in use: master @ 2483369355807050379e2385a4f22d16eb3437f7

BrianFreemanAtlanta:
Attempting to look into this one.  When I get into `AnalyzeConversion` the typeInfo/vbType = "ErrorType ?"  Not sure if additional information is needed on the conversion or add a case for the ErrorType or take another tack.
If you'd like me to do a pull request with the test case in it I can do that.

GrahamTheCoder:
If you see ErrorType or ErrorSymbol it generally means the compilation had errors.
If you look at `_semanticModel.GetDiagnostics()` you'll be able to see them.
In this case, adding an import should cover it:
```vbnet
Imports System.Collections

Public Module Module1
	Sub Main()
        Dim x As New ArrayList
        x.Add("a")

        Dim xs(1) As String

        xs(0) = x(0)

    End Sub
End Module
```

Other stuff:
* There's also a `_showCompilationErrors` flag I use for showing errors in the test output, but I recently broke it. It'll be fixed again as part of #60 
* Ideally that System.Collections import would have been automatically added for all conversions since it's a common one (and unused imports get removed at the end). Things like that should go into the list here: https://github.com/icsharpcode/CodeConverter/blob/master/ICSharpCode.CodeConverter/Util/VisualBasicCompiler.cs#L58 which if it "just works" would probably be a better fix rather than changing the test

GrahamTheCoder:
Generally it's also worth just having a solution with a VB and a CS project open alongside the code converter for checking syntax.

PR #390, opened 2019-10-17
Conversion for array index assignment

BrianFreemanAtlanta:
Link to issue(s) this covers
Closes Issue #365 
### Problem
Narrowing conversion from Object to string needed explicit ToString conversion added.

### Solution
* Added System.Collections to compilation options.  In second commit.  Can do separate PR if needed.
* Changed test to var declaration in C# instead of string declaration because it appears the converter prefers the var declaration.
* Added test case for simple object to string and more complex ListArray to string array assignment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to the code converter will be documented here.
 ### Vsix
 * Load extension only when menu item clicked (multi-project conversion menu not present until project loaded)
 
+### VB -> C#
+* Convert implicit object->string cast correctly (#365)
+
+### C# -> VB
+
+
 # 7.2.0 13/10/2019
 * Parallelize multi-file conversion
 * Make snippet conversion (ConvertText method) make classes partial by default since context isn't known (#379)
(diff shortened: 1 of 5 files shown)


GrahamTheCoder:
Great stuff. It didn't cause much faff this time, but to avoid conflicts in future, make sure you've fetched/pulled from origin (i.e. this repo), and start your branch from the latest master.
Thanks again!

mrmonday:
Thank you for doing this, @BrianFreemanAtlanta!

Couple of vaguely related things:

 * If you're doing the `CHANGELOG` thing, you might want to add [merge=union](https://www.git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-union) to the `.gitattributes` to avoid merge conflicts
 * If we have the `.vbproj`, we should pull the global import options from there (`<Import Include="System" />` etc) instead of hard-coding the list.

GrahamTheCoder:
Thanks @mrmonday . I'm totally on board with both of those. Please feel free to PR them.
If we made use of them from the compilation options here: https://github.com/icsharpcode/CodeConverter/blob/69fa87098ed194615e456ebe71f1d0704f9a17ee/ICSharpCode.CodeConverter/VB/NodesVisitor.cs#L105 then the hard-coded bit for when there's no vbproj could set them in compilation options rather than adding actual source text.

BrianFreemanAtlanta:
Will do. I started from the master I just missed merging to the latest version before starting the new branch after completing my other pull request.
Brian



Sent from my Verizon, Samsung Galaxy smartphone

Issue #92, opened 2018-04-15
C# -> VB: Properties with a modifer are being converted to auto properties


RussellSmith2:
### Input code
```C#
public string Text { get; private set; }
```

### Erroneous output
```vbnet
Public Property Text As String
```

### Expected output
```vbnet
Private _Text As String

Public Property Text As String
    Get
        Return _Text
    End Get
    Private Set(ByVal value As String)
        _Text = value
    End Set
End Property
```

### Details
Product in use: both

Version in use: 5.6.3

HasNoAccessorBody in VB\NodesVisitor.cs needs updated to check for modifiers on accessors.

BrianFreemanAtlanta:
I started to look into this issue.  There are 2 albeit related issues.
1. The stated `HasNoAccessorBody` corrections (and adding accessor/declaration for VB code)
2. Separate modifiers for get/set routines

Fixing task 1 the results will be:
```vb
Private _Text As String

Public Property Text As String
    Get
        Return _Text
    End Get
    Set(ByVal value As String)
        _Text = value
    End Set
End Property
```
With the missing modifier for the `Set` routine.
I looked at the reverse translation and the code doesn't handle the different modifiers from VB -> C# either.
Should I open a separate issue for task 2 and do a pull request on it before completing this?  Or see if a simple solution for task 2 can be done as part of this issue?

GrahamTheCoder:
I don't have good Internet access to look into the specific case right now,
but what you said makes sense. Feel free to combine related things together
if they are small and similar priority/risk to save bureacracy, so long as
the pr makes clear what it's trying to achieve. Try to at least keep them
in separate commits where possible so that if necessary they could be
easily separately merged.

GrahamTheCoder:
I think there may be another open issue to deal with modifiers more
generally using the CsSyntaxGenerator.

GrahamTheCoder:
In future we might consider specialising the case of a private setter since VB provides a more idiomatic way to write to the backing field of a readonly property. If anyone wants to open this as a separate enhancement, feel free, I'm just leaving it here now for completeness:

```csharp
public string Value {get; private set;}
```

```vbnet
Public ReadOnly Property Value As String
```

Within the class it's possible to use the private setter by using an underscore prefix: `_Value = "Fred"`

PR #402, opened 2019-10-29
Property with modifier

BrianFreemanAtlanta:
working on issue #92 .  This handles the modifiers for the property accessors.  The test is currently failing because the accessor bodies are still empty.  

### Problem

working on accessors with individual modifiers on them with no code body to convert.

### Solution
* I added a new tokenContext for Member in Property
* I'm stumbling a bit to create the local declaration and accessor bodies when needed.
* I added tests for both directions C# -> VB and VB -> C# for the accessor modifiers
* I added code in CommonConversion to add required declarations for property blocks similar to pattern expressions and some declaration statements
* I updated convertAccessor to fill in the accessor body when needed

Tasks for this PR:
- [x] Recognize and carry over accessor modifiers when needed
- [x] Fill in required body when needed

I ran all the tests and made corrections so the property block modifiers come over only when needed and the existing tests still pass.


GrahamTheCoder:
Looks like a good start. Have a look at the conversion in the other direction which probably has something similar. And try out roslyn quoter (link in the contributing.md file) if you haven't already for help generating specific structures. I'll try to add more specific tips if I get a minute later.

BrianFreemanAtlanta:
Just in case it didn't let you know.  I checked in an update that completes both part of the problem.  You can review and let me know if I need to make any changes.

GrahamTheCoder:
Sorry I couldn't be more help recently, but I've just now returned home to the land of Internet so I'll try to take a good look tomorrow. Great job making progress with minimal help!

GrahamTheCoder:
All looks good, I've just pushed a few minor tidyups. The one I'm least sure about was removing the test for the vb->cs direction. Let me know if that was supposed to stay. Thanks for your work!

BrianFreemanAtlanta:
Removing the extra test is fine.  I initially added it to confirm it was happening both directions, which it was.  That then confirmed for me that it was a full solution.

PR #406, opened 2019-11-11
VB -> C#: Support net standard projects

### LINKED ISSUE #398: VB -> C#: NullReferenceException when using the VS extension with .NET Standard
Hello,
I am trying to convert some VB.NET code to C# using the Visual Studio extension in VS2019. 
When I try to convert a file containing the input code below, an error message is shown:

CodeConvertError

No output file is generated. The same happens with other source files in the project as well, so it is probably no quirk of the code itself.

The solution contains both .NET Standard and .NET Framework projects, and weirdly this only occurs in .NET Standard (2.0) VB.NET projects. In .NET Framework both VB->C# and C#->VB works, in .NET Standard only C#->VB works.

I have no real clue why. I installed the extension by downloading the installer from the marketplace, instead of installing from the extension manager.
By virtue of the nice async/await stack traces I wasn't really able to find the code where it occurs :-)

Thanks for the work you put in!
Cheers
Jens

### Input code
```vbnet
Namespace CAD
    ''' <summary>
    ''' Represents a CAD RGB Color tuple
    ''' </summary>
    Public Structure CADColor
        Public Sub New(A As Byte, R As Byte, G As Byte, B As Byte)
            Me.A = A
            Me.R = R
            Me.G = G
            Me.B = B
        End Sub
        Public Sub New(R As Byte, G As Byte, B As Byte)
            Me.A = 255
            Me.R = R
            Me.G = G
            Me.B = B
        End Sub
        Public Property A As Byte
        Public Property R As Byte
        Public Property G As Byte
        Public Property B As Byte

        Public Shared Function FromArgb(R As Byte, G As Byte, B As Byte) As CADColor
            Return FromArgb(255, R, G, B)
        End Function
        Public Shared Function FromArgb(A As Byte, R As Byte, G As Byte, B As Byte) As CADColor
            Return New CADColor(A, R, G, B)
        End Function
    End Structure
End Namespace
```

### Details
* Version in use: 7.2.0.0
* Did you see it working in a previous version, which? --> Never used the extension before

GrahamTheCoder:
Fixes #398


TestData/CharacterizationTestSolution/VbNetStandardLib/VbNetStandardLib.vbproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>VbNetStandardLib</RootNamespace>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
GrahamTheCoder: Since this needs nuget packages to build it fails on the build server.
* [ ] Add code to the test to restore nuget packages so that we can properly test this

GrahamTheCoder:
This improves lots of things, but doesn't quite get the const field conversions right for Microsoft.VisualBasic.Constants in MyNamespace.Static.Designer.cs

I'll do some tweaks soon to neaten those up.

PR #407, opened 2019-11-11
VB -> C#: Fix implicit `ElementAtOrDefault` usage [#362]

### LINKED ISSUE #362: VB -> C#: Missing index for dictionary access
### Input code
```vbnet
Imports System
Imports System.Collections.Generic

Module Module1
    Dim Dict As New Dictionary(Of Integer, String)

    Sub Main()
        Dim x = Dict.Values(0).Length
    End Sub
End Module
```

### Erroneous output
```c#

namespace ConsoleApp10
{
    static class Module1
    {
        private static Dictionary<int, string> Dict = new Dictionary<int, string>();

        public static void Main()
        {
            var x = Dict.Values.Length;
        }
    }
}
```

### Expected output
```c#

namespace ConsoleApp10
{
    static class Module1
    {
        private static Dictionary<int, string> Dict = new Dictionary<int, string>();

        public static void Main()
        {
            var x = Dict.Values[0].Length;
        }
    }
}

```

### Details

Version in use: master @ 2483369355807050379e2385a4f22d16eb3437f7

This case is quite interesting, since it's still valid code - it just does the wrong thing. My original case was with a dictionary of some other class, with some other property, so the code did not compile.

mrmonday:
Ensure parameterless property access short-circuit only happens when there are no parameters. This allows the existing code for `ElementAtOrDefault` to be triggered correctly.

Fixes #362

### Problem

The shortcut for property access caused the `ElementAtOrDefault` handling to not trigger.

### Solution

Checking the parameter length first allows this to work as expected.

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -651,7 +651,7 @@ namespace ICSharpCode.CodeConverter.CSharp
                 return await CreateElementAccess();
             }
 
-            if (expressionSymbol != null && expressionSymbol.IsKind(SymbolKind.Property)) {
+            if (expressionSymbol != null && expressionSymbol.IsKind(SymbolKind.Property) && invocationSymbol.GetParameters().Length == 0) {
                 return convertedExpression; //Parameterless property access
             }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to the code converter will be documented here.
 * Convert trivia (e.g. comments) at start of file (#333)
 * Improvements to redim conversion (#403, #393)
 * Convert array of arrays initializer (#364)
+* Convert implicit `ElementAtOrDefault` (#362)
 
 ### C# -> VB
 * Convert property accessors with visiblity modifiers (#92)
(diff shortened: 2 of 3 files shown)


ICSharpCode.CodeConverter/CSharp/ExpressionNodeVisitor.cs
```diff
@@ -651,7 +651,7 @@ public override async Task<CSharpSyntaxNode> VisitInvocationExpression(VBasic.Sy
                 return await CreateElementAccess();
             }
 
-            if (expressionSymbol != null && expressionSymbol.IsKind(SymbolKind.Property)) {
+            if (expressionSymbol != null && expressionSymbol.IsKind(SymbolKind.Property) && invocationSymbol.GetParameters().Length == 0) {
```
GrahamTheCoder: I have a nagging feeling that there's a more general case that may still be missed. I wonder whether this if statement should be below the other one. I'll merge this since it's definitely better than it was, just wondering if you have any thoughts on more general cases - I'll post here if I think of a specific case where it's still broken.

mrmonday: I can't think of anything off the top of my head. I think you're probably right about the ordering - though unless we think of another case I don't think it currently makes a difference.

GrahamTheCoder: Yep, I've thought about it a bit more and they seem mutually exclusive so I think this covers it, thanks!

PR #412, opened 2019-11-11
Convert short literals in ternary expressions [#363]

### LINKED ISSUE #363: VB -> C#: Missing conversion for short literal
### Input code
```vbnet
Module Module1
    Sub Main()
        Dim x As Short = If(True, CShort(50), 100S)
    End Sub
End Module
```

### Erroneous output
```c#
namespace ConsoleApp10
{
    static class Module1
    {
        public static void Main()
        {
            short x = true ? Conversions.ToShort(50) : 100;
        }
    }
}
```

### Expected output
```c#
namespace ConsoleApp10
{
    static class Module1
    {
        public static void Main()
        {
            short x = Conversions.ToShort(true ? 50 : 100);
        }
    }
}
```

### Details

Version in use: master @ 2483369355807050379e2385a4f22d16eb3437f7

Gives [CS0266](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0266)

mrmonday:
* Add explicit conversions for both branches of a ternary expression
 * Add conversions for short literals

Fixes #363

### Problem

Short literals in ternary expressions were not converted.

### Solution

There were a couple of things here - we need to add conversions for each branch of the ternary opeartor. We could probably be more clever about this, but just adding the conversion is easy enough.

The other thing is the conversion of VB short literals. I added a case to the conversion analysis code to handle this - another (more invasive) approach would be to add an explicit case whenever we convert a VB literal that has an explicit suffix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to the code converter will be documented here.
 * Improvements to implicit enum -> int conversion (#361)
 * Convert expressions in constants (#329)
 * Convert implicit `ElementAtOrDefault` (#362)
+* Convert types in ternary expressions (#363)
 
 ### C# -> VB
 * Convert property accessors with visiblity modifiers (#92)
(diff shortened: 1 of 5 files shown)


mrmonday:
Not sure why CI is failing - running the failing tests locally seems to work.

GrahamTheCoder:
I reran and it passed, and it works locally too. I recently allowed files to be converted in parallel, so it just happened to show up this race condition. It wasn't anything to do with your PR, but since we saw it here and it's a tiny fix I've added it on the end.

PR #413, opened 2019-11-11
VB -> C#: Duplicate cases Should be Discarded

### LINKED ISSUE #374: VB -> C#: Duplicate cases should be discarded
VB allows `Case` in a `Select Case` to have duplicate labels.

### Input code
```vbnet
Imports System

Friend Module Module1
    Sub Main()
        Select Case 1
            Case 1
                Console.WriteLine("a")

            Case 1
                Console.WriteLine("b")

        End Select
        Console.ReadLine()
    End Sub
End Module
```

This compiles, and outputs `a`.

### Erroneous output

```c#
using System;

namespace ConsoleApp3
{
    internal static class Module1
    {
        public static void Main()
        {
            switch (1)
            {
                case 1:
                    {
                        Console.WriteLine("a");
                        break;
                    }

                case 1:
                    {
                        Console.WriteLine("b");
                        break;
                    }
            }
            Console.ReadLine();
        }
    }
}
```

### Expected output
```c#
using System;

namespace ConsoleApp3
{
    internal static class Module1
    {
        public static void Main()
        {
            switch (1)
            {
                case 1:
                    {
                        Console.WriteLine("a");
                        break;
                    }
            }
            Console.ReadLine();
        }
    }
}
```

### Details

Error:
```
error CS0152: The switch statement contains multiple cases with the label value '1'
```

I wonder if it's worth leaving the code there, but commented. Or even leaving the error - the behaviour changes, since it won't compile, but I can't think of a case where this wouldn't be a bug.

Version: master @ 3ffc597ffdce991ef0cccf01bd7c884b9d337d6a

BrianFreemanAtlanta:
Link to issue(s) this covers
this fixes #374.
### Problem
Duplicate case statements resulted in compile error

### Solution
* Added warning stating the case was unreachable.  Display converted code to aid in debugging if needed.
* Instead of hard coding this message I added extensions like the `WithCsTrailingErrorComment` and `DescribeConversionError`

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/MethodBodyExecutableStatementVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/MethodBodyExecutableStatementVisitor.cs
@@ -553,7 +553,6 @@ namespace ICSharpCode.CodeConverter.CSharp
             var exprWithoutTrivia = expr.WithoutTrivia().WithoutAnnotations();
             var usedConstantValues = new HashSet<object>();
             var sections = new List<SwitchSectionSyntax>();
-            var cases = new List<String>();
             foreach (var block in node.CaseBlocks) {
                 var labels = new List<SwitchLabelSyntax>();
                 foreach (var c in block.CaseStatement.Cases) {
--- a/ICSharpCode.CodeConverter/CSharp/MethodBodyExecutableStatementVisitor.cs
+++ b/ICSharpCode.CodeConverter/CSharp/MethodBodyExecutableStatementVisitor.cs
@@ -597,7 +597,7 @@ namespace ICSharpCode.CodeConverter.CSharp
                     //Else write comment to code?
                     var lastCase = sections.Last();
                     sections.Remove(lastCase);
-                    lastCase = lastCase.WithCsTrailingWarningComment(block, new NotSupportedException("Duplicate Case label commented out"));
+                    lastCase = lastCase.WithCsTrailingWarningComment("Unreachable case label","This case was not executable prior to code conversion, but is preserved here in case it helps troubleshoot a bug", SyntaxFactory.SwitchSection(SyntaxFactory.List(labels), list));
                     sections.Add(lastCase);
                 }
             }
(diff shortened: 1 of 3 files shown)


ICSharpCode.CodeConverter/CSharp/MethodBodyExecutableStatementVisitor.cs
```diff
@@ -587,7 +588,18 @@ public override async Task<SyntaxList<StatementSyntax>> VisitSelectBlock(VBSynta
                     csBlockStatements.Add(SyntaxFactory.BreakStatement());
                 }
                 var list = SingleStatement(SyntaxFactory.Block(csBlockStatements));
-                sections.Add(SyntaxFactory.SwitchSection(SyntaxFactory.List(labels), list));
+                var labelDesc = string.Join(" ", labels.Select(l => l.ToFullString()));
```
GrahamTheCoder: ToFullString would include trivia, ToString would be better. Though technically there are several ways of writing equivalent cases, such as using hex, binary or octal representations of numbers. We only need to deal with constants for the sake of our purposes here (since others will become pattern match cases), so using _semanticModel.GetConstantValue might be best.

ICSharpCode.CodeConverter/CSharp/MethodBodyExecutableStatementVisitor.cs
```diff
@@ -587,7 +588,18 @@ public override async Task<SyntaxList<StatementSyntax>> VisitSelectBlock(VBSynta
...
+                var numMatch = cases.Count(c => c == labelDesc);
+                if (numMatch == 0) {
+                    cases.Add(labelDesc);
+                    sections.Add(SyntaxFactory.SwitchSection(SyntaxFactory.List(labels), list));
+                } else {
+                    //Else write comment to code?
+                    var lastCase = sections.Last();
```
GrahamTheCoder: If there were 3 duplicates, we'd actually want to comment out all except the first

GrahamTheCoder:
Thanks for this. I've pushed some changes to fix the edge cases mentioned and I've moved it to using a hashset rather than checking through existing cases in each loop. Then I found that using the pattern matching solution gets the compiler to do the warning for us so I've flipped it to using that.

PR #414, opened 2019-11-12
Correctly handle type promoted module symbols [#332, #375, #401]

### LINKED ISSUE #332: VB -> C#: Ambiguous reference for names in automatic imports
### Input code
File 1:
```vbnet
Namespace A
    Public Class Dictionary(Of V, K)
        Inherits System.Collections.Generic.Dictionary(Of K, V)
        Public Sub New()
        End Sub
    End Class
End Namespace
```
File 2:
```vbnet
Imports ConsoleApp4.A

Friend Class Program
    Public Shared Function Foo() As System.Collections.Generic.Dictionary(Of String, String)
        Return New Dictionary(Of String, String)()
    End Function

    Public Shared Sub Main(ByVal args As String())
        Dim x As List(Of String) = New List(Of String)()
    End Sub
End Class
```

### Erroneous output
File 1:
```c#
namespace ConsoleApp4
{
    namespace A
    {
        public class Dictionary<V, K> : System.Collections.Generic.Dictionary<K, V>
        {
            public Dictionary()
            {
            }
        }
    }
}
```

File 2:
```c#
using ConsoleApp4.A;

namespace ConsoleApp4
{
    internal class Program
    {
        public static System.Collections.Generic.Dictionary<string, string> Foo()
        {
            return new Dictionary<string, string>();
        }

        public static void Main(string[] args)
        {
            List<string> x = new List<string>();
        }
    }
}
```

Error:

```
error CS0104: 'Dictionary<,>' is an ambiguous reference between 'ConsoleApp4.A.Dictionary<V, K>' and 'System.Collections.Generic.Dictionary<TKey, TValue>'
```

### Expected output
File 1 - the same.

File 2:

```c#
using ConsoleApp4.A;

namespace ConsoleApp4
{
    internal class Program
    {
        public static System.Collections.Generic.Dictionary<string, string> Foo()
        {
            return new A.Dictionary<string, string>();
        }

        public static void Main(string[] args)
        {
            List<string> x = new List<string>();
        }
    }
}
```

The returned dictionary in `Foo` is explicitly qualified here. There are a few other ways to resolve this too.

### Details
Product in use: VS extension

Version in use: 6.9.0.0

### LINKED ISSUE #375: VB -> C#: Explicitly qualified global names not translated
For names which can be used globally, eg. `Timer`, `Now`, `Len`, etc, VB allows you to explicitly qualify them as eg. `Microsoft.VisualBasic.Timer`. The actual symbol is elsewhere, eg. `Microsoft.VisualBasic.DateAndTime.Timer`. This is not converted correctly.

### Input code
```vbnet
Friend Module Module1
    Sub Main()
        Dim x = Microsoft.VisualBasic.Timer
    End Sub
End Module
```

### Erroneous output
```c#
namespace ConsoleApp3
{
    internal static class Module1
    {
        public static void Main()
        {
            double x = Microsoft.VisualBasic.Timer;
        }
    }
}
```

### Expected output
```c#
using Microsoft.VisualBasic;

namespace ConsoleApp3
{
    internal static class Module1
    {
        public static void Main()
        {
            double x = DateAndTime.Timer;
        }
    }
}
```

### Details

Error:
```
error CS0234: The type or namespace name 'Timer' does not exist in the namespace 'Microsoft.VisualBasic' (are you missing an assembly reference?)
```

Version: master @ 3ffc597ffdce991ef0cccf01bd7c884b9d337d6a

### LINKED ISSUE #401: VB -> C#: deal with using alias in imports : Imports VB = Microsoft.VisualBasic
### Input code
```vbnet
Imports VB = Microsoft.VisualBasic
...
copy = VB.Left(sometext, i_1)
```
```C#
If relevant, please enter some example code here to reproduce the issue (in VB or C#), or the steps taken to cause the issue.
```

### Erroneous output
```C#
copy = Microsoft.VB.Left(sometext, i_1);
			
```

### Expected output
```C#
copy = Microsoft.VisualBasic.Left(sometext, i_1); // which would be wrong
...or better...
copy =Microsoft.VisualBasic.Strings.Left(sometext, i_1);  
...or...
using VB= Microsoft.VisualBasic;
copy =VB.Strings.Left(sometext, i_1);  
```

### Details
* Version in use: 7.2.0.0

mrmonday:
Fixes #332
Fixes #375
Fixes #401

### Problem

We get the incorrect type when analysing type promoted module symbols, causing the incorrect symbol to be used when qualifying names explicitly.

Additionally, renamed imports are removed when they match an import that's added automatically.

When an unqualified name is used in VB, the converted C# did not qualify it if it would become ambiguous.

### Solution

Use the containing type of the symbol to determine the correct LHS for a `MemberAccessExpression`. This conveniently handles type promotion, so we can remove the special case.

If there are multiple imports of the same namespace, prefer the import with an alias, and use that for name resolution.

When converting a generic name, qualify it if possible to ensure we select the right reference.


ICSharpCode.CodeConverter/CSharp/DeclarationNodeVisitor.cs
```diff
@@ -99,8 +99,8 @@ public override async Task<CSharpSyntaxNode> VisitCompilationUnit(VBSyntax.Compi
                 .SelectAsync(async c => (UsingDirectiveSyntax) await c.Accept(TriviaConvertingVisitor));
             var usingDirectiveSyntax = usings
                 .Concat(_extraUsingDirectives.Select(u => SyntaxFactory.UsingDirective(SyntaxFactory.ParseName(u))))
-                .GroupBy(u => u.ToString())
-                .Select(g => g.First());
+                .GroupBy(u => u.Name.ToString())
+                .Select(g => g.OrderBy(v => !string.IsNullOrEmpty(v.Alias?.ToString()) ? 0 : 1).First());
```
GrahamTheCoder: I think we should be grouping on name *and* alias. It's possible to alias the same import as many different things

ICSharpCode.CodeConverter/CSharp/ExpressionNodeVisitor.cs
```diff
@@ -990,23 +987,6 @@ private ExpressionSyntax GetConvertMethodForKeywordOrNull(SyntaxNode type)
...
-                    nameContainingSymbol &&
-                nameContainingSymbol.ContainingSymbol.Equals(expressionSymbol)) {
-                moduleSymbol = nameContainingSymbol;
-                return true;
-            }
-
-            moduleSymbol = null;
-            return false;
-        }
-
```
GrahamTheCoder: Nice!

ICSharpCode.CodeConverter/CSharp/DeclarationNodeVisitor.cs
GrahamTheCoder: I'll tweak and merge this now

mrmonday: I think it's a little more nuanced than that, now you mention it. I'm not sure what the right fix is without a bit more thought.

My thought behind this implementation was "if we specified an alias for a name, prefer the alias". I hadn't considered multiple aliases for the same name.

The previous implementation grouped by name, so after simplification would never use the alias (there would be two `using`s, and the aliased one would be simplified away).

GrahamTheCoder: When testing this I came across a case that doesn't work. I've pushed the failing test and the tweak mentioned. I'll leave it as is to give you a chance to have a look at it, feel free to nuke my changes as always if you have a better plan

GrahamTheCoder: Ah, I hadn't seen your comment when you posted. I've contrived the test to avoid the simplifier getting involved.

mrmonday: Ended up down a bit of a rabbit hole here.

 * The test case you added doesn't compile in VB
 * I hit a few unrelated bugs along the way while making it compile
 * The name is unambigous when unqualified in C#, since types and names are in different namespaces

I've pushed up some changes which seem to handle this properly. I'll put together some test cases and file bugs for the things I hit along the way.

mrmonday: Filed #418 for the issues I found along the way. Think this PR should be good now, unless you can think of some more unhandled cases.

mrmonday:
Looks like the other PR accidentally got merged in here... I'll clean it up and close the other PR - they touch similar code, so it's worth doing them together.

mrmonday:
Rebased and merged everything into this single PR - I won't touch it again until I hear back from you.

I've left the clean-up commits mostly in tact, I could squash them into the relevant other commits though if you'd prefer.

mrmonday:
Looks like this has some issues with converting things in the global namespace - I'll see if I can narrow down a test case and take a look at it tomorrow.

GrahamTheCoder:
Yep that easily surpasses my standards for git history tidiness, no need to do any more tidying.
I see green tests, nice one! Yep global case makes sense to get in this PR too, go for it

mrmonday:
Global case is now handled, this should be good to go.

PR #415, opened 2019-11-12
Prefer renamed imports for symbol resolution [#401]

### LINKED ISSUE #401: VB -> C#: deal with using alias in imports : Imports VB = Microsoft.VisualBasic

mrmonday:
Fixes #401 

### Problem

Renamed imports are removed when they match an import that's added automatically.

### Solution

If there are multiple imports of the same namespace, prefer the import with an alias, and use that for name resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to the code converter will be documented here.
 * Convert types in ternary expressions (#363)
 * Support for converting dot net standard VB projects (#398)
 * Avoid compilation error for duplicate cases (#374)
+* Prefer renamed imports for name resolution (#401)
 
 ### C# -> VB
 * Convert property accessors with visiblity modifiers (#92)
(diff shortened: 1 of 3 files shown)


mrmonday:
Oooh, ok, this actually depends on #414. I'll close this and merge the changes in.

PR #424, opened 2019-11-14
Ensure casing matches for namespaces [#421]

### LINKED ISSUE #421: VB -> C#: Namespaces with different cases merged
### Input code
```vbnet
Namespace [Aaa]
    Friend Class A
    End Class
End Namespace

Namespace Global.aaa
    Friend Class B
    End Class
End Namespace

Friend Module C
    Sub Main()
        Dim x = New aaa.A
        Dim y = New Aaa.B
    End Sub
End Module
```

### Erroneous output
```c#
namespace Aaa
{
    internal class A
    {
    }
}

namespace Aaa
{
    internal class B
    {
    }
}

internal static class C
{
    public static void Main()
    {
        var x = new Aaa.A();
        var y = new Aaa.B();
    }
}
```

### Expected output
```c#
namespace Aaa
{
    internal class A
    {
    }
}

namespace aaa
{
    internal class B
    {
    }
}

internal static class C
{
    public static void Main()
    {
        var x = new Aaa.A();
        var y = new aaa.B();
    }
}
```

### Details

* Version in use: 7.3.0

The original VB is hard to enter in VS, since it likes to auto-correct. The original case was cross-project, where it's easier to achieve (but harder to put into a test case). According to ILSpy this does end up as two different namespaces.

mrmonday:
Fixes #421

### Problem

When converting multiple namespaces with the same name but different cases, the symbols from the semantic info get merged, despite being distinct symbols. Roslyn recognises this internally (using a `MergedNamespaceDeclaration`), but doesn't seem to expose it.

When this happens, `CsSyntaxGenerator` will always use the casing of the first namespace declaration it encountered, even if this is incorrect.

### Solution

 * When converting identifiers, only update the case if this is *not* a declaration
 * When getting a `TypeSyntax` for conversion, manually replace the identifiers when the strings are identical, but cases don't match.

The tests for this will fail until #414 is merged. I didn't want to include it in there, since it's already doing quite a lot, and my solution to this seems rather hacky.

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
+++ b/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
@@ -170,6 +170,11 @@ namespace ICSharpCode.CodeConverter.CSharp
             return WithDeclarationCasing(syntax, typeSymbol);
         }
 
+        /// <summary>
+        /// Semantic model merges the symbols, but the compiled form retains multiple namespaces, which (when referenced from C#) need to keep the correct casing.
+        /// <seealso cref="DeclarationNodeVisitor.WithDeclarationCasing(VBSyntax.NamespaceBlockSyntax, ISymbol)"/>
+        /// <seealso cref="CommonConversions.WithDeclarationCasing(SyntaxToken, ISymbol, string)"/>
+        /// </summary>
         private static TypeSyntax WithDeclarationCasing(TypeSyntax syntax, ITypeSymbol typeSymbol)
         {
             var vbType = SyntaxFactory.ParseTypeName(typeSymbol.ToDisplayString());
@@ -256,6 +261,11 @@ namespace ICSharpCode.CodeConverter.CSharp
             return CsEscapedIdentifier(text);
         }
 
+        /// <summary>
+        /// Semantic model merges the symbols, but the compiled form retains multiple namespaces, which (when referenced from C#) need to keep the correct casing.
+        /// <seealso cref="DeclarationNodeVisitor.WithDeclarationCasing(VBSyntax.NamespaceBlockSyntax, ISymbol)"/>
+        /// <seealso cref="CommonConversions.WithDeclarationCasing(TypeSyntax, ITypeSymbol)"/>
+        /// </summary>
         private static string WithDeclarationCasing(SyntaxToken id, ISymbol symbol, string text)
         {
             bool isDeclaration = symbol.Locations.Any(l => l.SourceSpan == id.Span);
--- a/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
+++ b/ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
@@ -165,16 +165,21 @@ namespace ICSharpCode.CodeConverter.CSharp
         public TypeSyntax GetTypeSyntax(ITypeSymbol typeSymbol, bool useImplicitType = false)
         {
             if (useImplicitType || typeSymbol == null) return CreateVarTypeName();
-            var vbType = SyntaxFactory.ParseTypeName(typeSymbol.ToDisplayString());
-            var originalNames = vbType.DescendantNodes().OfType<CSSyntax.IdentifierNameSyntax>().Select(i => i.ToString()).ToList();
-            var syntax = (TypeSyntax) CsSyntaxGenerator.TypeExpression(typeSymbol);
+            var syntax = (TypeSyntax)CsSyntaxGenerator.TypeExpression(typeSymbol);
 
-            return syntax.ReplaceNodes(syntax.DescendantNodes().OfType<CSSyntax.IdentifierNameSyntax>(), (oldNode, n) => {
-                var originalName = originalNames.Where(on => string.Equals(on, oldNode.ToString(), StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
-                if (originalName != null) {
-                    return SyntaxFactory.IdentifierName(originalName);
-                }
-                return oldNode;
+            return WithDeclarationCasing(syntax, typeSymbol);
+        }
+
+        private static TypeSyntax WithDeclarationCasing(TypeSyntax syntax, ITypeSymbol typeSymbol)
+        {
+            var vbType = SyntaxFactory.ParseTypeName(typeSymbol.ToDisplayString());
+            var originalNames = vbType.DescendantNodes().OfType<CSSyntax.IdentifierNameSyntax>()
+                .Select(i => i.ToString()).ToList();
+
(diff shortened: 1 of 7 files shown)


ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
```diff
@@ -165,7 +165,17 @@ private TypeSyntax GetFuncTypeSyntax(IMethodSymbol method)
         public TypeSyntax GetTypeSyntax(ITypeSymbol typeSymbol, bool useImplicitType = false)
         {
             if (useImplicitType || typeSymbol == null) return CreateVarTypeName();
-            return (TypeSyntax) CsSyntaxGenerator.TypeExpression(typeSymbol);
+            var vbType = SyntaxFactory.ParseTypeName(typeSymbol.ToDisplayString());
```
GrahamTheCoder: I expect a VbSyntaxGenerator would do this correctly. Might be worth pulling one through and doing the logic on a generated vb tree rather than parsing typeSymbol.ToDisplayString() which might have a few unexpected quirks.

GrahamTheCoder:
Good work on a very gnarly issue. I've tried to pick out some edge cases in the inline comments. Since this is something complicated, it'd be good to pull the logic out into a method and give it a name where possible, so someone can connect these 3 bits of code together. e.g. "WithNamespaceDeclaredCasing" (and then put xmldoc on that method saying "See WithIdentifierDeclaredCasing" along with some of the description from this PR).

Since they're separate namespaces, is it legitimate in vb to define a class `Aaa.C` *and* `aaa.c`?
It might be worth using a selfverifying test or two for some of these since they're weird enough that someone might later come along and change the expected result without understanding the full complexity.

mrmonday:
There's lots of overlap, so going to attempt to address everything in one comment.

> Good work on a very gnarly issue.

Thanks... It was fiddly, and as you've seen, my solution isn't ideal.

> Since they're separate namespaces, is it legitimate in vb to define a class `Aaa.C` _and_ `aaa.c`?

As far as I can tell, it isn't. Given `Aaa.A` and `aaa.B`, you get a warning and `Aaa.A`, and `aaa.B` in the assembly. For `Aaa.C` and `aaa.c` you get an error about conflicting definitions.

This makes it particularly hard to test - it's completely legitimate to reference other projects or assemblies which contain any amount of casing combinations, and you will be able to reference them from VB.

> In the problem/issue descriptions it only talks about namespaces, but this bit seems like it'll have the same effect on other identifiers.

This is correct - definitely a bug. I need to check the behaviour for partial classes (I suspect it merges them). For methods we dealt with this previously - it uses the case of the method in the base class/interface.

> I expect a VbSyntaxGenerator would do this correctly

I intentionally didn't use this, since I don't know what it does - if you parse the syntax you get exactly the tokens you put in, without the case changing. I'll take a look at it and see how it goes.

> Probably better off either fully parsing, or splitting on dots at least

I don't think it matters here - there shouldn't be a situation where there's a mismatch like you describe, since both methods parse the same namespace, and we only do the replacement when the suffixes match. I can't think of a scenario where visiting would give `Blah` and the symbol would give `A.FooBlah`.

> it'd be good to pull the logic out into a method

Agreed, I'll do that.

> It might be worth using a selfverifying test

That's a good idea, I'll move the test case I think.

ICSharpCode.CodeConverter/CSharp/CommonConversions.cs
mrmonday: Tried this, and it has the same problem as `CsSyntaxGenerator` - it uses the wrong casing.

mrmonday:
I've added a couple of extra tests in there for the cases you pointed out. I haven't factored anything out into methods, I wasn't sure where to draw the line or what to call the methods.

mrmonday:
Just tested this against my original project - looks like there's still more work to do here, I'll see if I can narrow it down.

GrahamTheCoder:
OK, let me know when you're happy with the PR. At that point I'll:
* [ ] Attempt to pull out the logic at into methods at the end. I suspect they will feel a bit clunky but hopefully it'll be worth it.
* [ ] Have another think through and write a test for any situations I'm still not sure of e.g. I need to revisit the `A.FooBlah` situation

mrmonday:
I am "happy with it" in that more tests pass than before, and I don't see any regressions on the real world project I'm working against.

I'm struggling to build a test case for the real-world case I have - visual studio keeps crashing while I debug the vsix (outside of code converter), which makes it very hard to figure hit a breakpoint where I need it. I'll keep poking it.

Relatedly, is there a way to set the maximum parallelism for the converter to `1` so I don't have to deal with context switching while debugging?

mrmonday:
> I am "happy with it"

By this I mean, I'm happy for you to take over - I don't plan on any more tweaks to what's there, and anything further I do I can do on top of your changes.

GrahamTheCoder:
Great, I'll have a look. You can hardcode Env.MaxDop to 1 to avoid context switching.
It's worth checking the [activity log](https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-use-the-activity-log?view=vs-2019) of visual studio for a stacktrace too
So long as it only occurs when the converter is running, feel free to open an issue for it even if you can't pin down the details just in case someone else has another piece of the puzzle.
If you are getting a stacktrace of any kind (even outside the converter) stick it in the issue and it might be possible to take some guesses at what could cause it.

GrahamTheCoder:
In fact, I've just pushed a change to master so that if there's a debugger attached in debug mode it defaults to 1. This does mean of course that if there's a parallelism specific issue we'll need to remember to tweak that.

mrmonday:
Thanks for the `Env.MaxDop` tip - that was all I needed to narrow it down.

I haven't made a test case, but it should just need a variant of:

```csharp
A.B.SharedMethod();
```

To reproduce the issue I was encountering.

Changing:
https://github.com/icsharpcode/CodeConverter/blob/81f9aa2465912e9bbfe83db9a843d59c4308e171/ICSharpCode.CodeConverter/CSharp/ExpressionNodeVisitor.cs#L286

To:
```csharp
if (type != null) {
```

Fixes the other case that I mentioned. Since `GetTypeSyntax()` was never called to get the type for `MemberAccessExpression`s, it ended up in `ConvertIdentifier`, and used the source casing, rather than the symbol casing.

I'll leave it to you to do this - don't want to tred on your toes while you're working on it.

GrahamTheCoder:
I'm done making changes, I'll hand back over to you for that last case since I've not fully understood it.
Make sure to checkout the latest since I rebased it.

mrmonday:
I've added/fixed the test case I mentioned. Should be good to go now!

Thanks for working on this with me 🙂

PR #434, opened 2019-11-20
C# -> VB fix some indexer's bugs, escaping variables with name of predefined types, improve base type's resolving

Alexilia:
Link to issue(s) this covers

### Problem
All changes are for C# -> VB conversion

1. Converting of variable of predefined types like DateTime should be escaped, when variable's name the same as type's name (DateTime date -> Date [date])
2. Explicitly implemented interface members doesn't contains Implements keyword
3. Simplifier.ReduceAsync sometimes with null refs, so catch it to leave document as is.

4. Indexer with empty setter is applicable, and no need in backing field,  when the getter is not empty.
5. Private Default members is not compiled. It is actually for indexers which implements IList, for example
6. Implements IList.this[]
 instead of 
Implements IList.Item
7. GetVbPropertyBackingFieldName(parent) fails on casting BasePropertyDeclarationSyntax to PropertyDeclarationSyntax
theres is no need to call the method for indexers, so test it to avoid exceptions in the future

### Solution

Tests for all cases in the PR


ICSharpCode.CodeConverter/CSharp/DocumentExtensions.cs
```diff
@@ -21,7 +21,11 @@ public static async Task<Document> WithSimplifiedSyntaxRootAsync(this Document d
         {
             var root = syntaxRoot  ?? await doc.GetSyntaxRootAsync();
             var withSyntaxRoot = doc.WithSyntaxRoot(root.WithAdditionalAnnotations(Simplifier.Annotation));
-            return await Simplifier.ReduceAsync(withSyntaxRoot);
+            try {
+                return await Simplifier.ReduceAsync(withSyntaxRoot);
+            } catch {
+                return doc;
+            }
```
GrahamTheCoder: This probably means some of our returned syntax trees are malformed. The only way we'll find out is if people report it as a bug that code isn't as simple as it could be. I'll merge as is, but it'd be good to get this logged to the output window via iprogress or something

GrahamTheCoder:
Great, thanks!

PR #438, opened 2019-11-23
Object initializers, nullables and awaits

GrahamTheCoder:
* [x] Exit Function
* [x] Imitate what CType(blahString, MyEnum) does in IL
* [x] Object initializer should apply type conversions (e.g. enum to int)
* [x] Now should go to DateAndTime.Now
* [x] Ensure async keyword transferred on async lambdas
* [x] If statement should be coalesced from nullable to bool. Check what "Not" operator does to a nullable bool

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,23 @@ All notable changes to the code converter will be documented here.
 * Correctly convert ambiguous names ([#332](https://github.com/icsharpcode/CodeConverter/pull/332))
 * Ensure correct visibility for constructors ([#422](https://github.com/icsharpcode/CodeConverter/pull/422))
 * Ensure casing is correct for namespaces ([#421](https://github.com/icsharpcode/CodeConverter/pull/421))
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,26 @@ All notable changes to the code converter will be documented here.
 * Correctly convert ambiguous names ([#332](https://github.com/icsharpcode/CodeConverter/pull/332))
 * Ensure correct visibility for constructors ([#422](https://github.com/icsharpcode/CodeConverter/pull/422))
 * Ensure casing is correct for namespaces ([#421](https://github.com/icsharpcode/CodeConverter/pull/421))
+* Convert implicit object->string cast correctly (#365)
```
mrmonday: Think you found the downside to `merge=union` here, @GrahamTheCoder 😉

GrahamTheCoder: Ah yes. I'll fix it on master

Issue #435, opened 2019-11-21
C# -> VB: wrong conversion of Custom event


Alexilia:
### Input code
```C#
using System;

class TestClass
{
    EventHandler backingField;

    public event EventHandler MyEvent {
        add {
            this.backingField += value;
        }
        remove {
            this.backingField -= value;
        }
    }
}
```

### Erroneous output
```vbnet
Imports System

Friend Class TestClass
    Private backingField As EventHandler

    Public Event MyEvent As EventHandler
        AddHandler(ByVal value As EventHandler)
            AddHandler backingField, value
        End AddHandler
        RemoveHandler(ByVal value As EventHandler)
            RemoveHandler backingField, value
        End RemoveHandler
    End Event
End Class
```

### Expected output
```vbnet
Imports System

Friend Class TestClass
    Private Event backingField As EventHandler

    Public Custom Event MyEvent As EventHandler
        AddHandler(ByVal value As EventHandler)
            AddHandler backingField, value
        End AddHandler
        RemoveHandler(ByVal value As EventHandler)
            RemoveHandler backingField, value
        End RemoveHandler
        RaiseEvent(ByVal sender As System.Object, ByVal e As EventArgs)
            RaiseEvent backingField(sender, e)
        End RaiseEvent
    End Event
End Class
```

### Details
wrong test CodeConverter.Tests.VB.MemberTest.TestCustomEvent()
it lacks:
Custom modifiers
Event keyword
RaiseEvent statement

PR #442, opened 2019-11-25
c# -> vb custom event improvements

Alexilia:
fix conversion of custom event with backing field

### Problem
https://github.com/icsharpcode/CodeConverter/issues/435


GrahamTheCoder:
Thanks. I'm not quite sure whether it'll do any harm marking all private delegates as events. At some point we might need to do something cleverer, but this seems like a great first step.

If you're planning to contribute quite a bit of C# to VB conversion code, it's probably worth being aware of https://github.com/icsharpcode/CodeConverter/issues/8#issuecomment-554673493
Paul converted the converter to VB, then worked on some improvements. At some point I was intending to convert his code back to C# and diff them (which will be quite messy) to see what could be merged back. I haven't got around to it yet, but it's possibly a good place to look for hints at minimum.

Alexilia:
Merry christmas! Next week I will prepare a fix for redundant event keyword, and if I find more common case to remove parentheses, I will add this too.

PR #449, opened 2019-11-28
Improve and move QualifyNode

GrahamTheCoder:
`QualifyNode` struggles in some more extreme cases where it'll accidentally qualify an instance reference with a static class. I have some examples I'll add as tests on here.

### Implementation
(On a now parked branch) I started trying to improve QualifyNode my making use of `IInstanceOperation`, but it involved understanding a serious number of possible cases. These are all cases which are built into Simplifier.Expand. Therefore I'm taking an iterative approach to reimplementing what I previously reverted here: https://github.com/icsharpcode/CodeConverter/commit/1780b71e240867a642ce4f20025293160d95ec31

The idea is to move the conditional code that decides whether to run QualifyNode to an up front step, and run Expand instead. The benefit here, is that we can apply the logic to all nodes in a single pass, then get a semantic model for the rewritten tree. This avoids CPU/memory issues of switching out the semantic model mid-conversion as well as keeping the code complexity out of the way.

### Details

You can't trust the syntax node type and structure since missing argument lists aren't understood at a parse level, and VB allows type promotion.
Hence we must look at an operation which can distinguish the syntax to access an instance, from the symbol it represents.

An expression is formed of three parts in one of the following patterns:
* InstanceHolder (e.g. MyStaticPropertyHoldingAnInstance or new List(Of String) or Me or MyInScopeVar or MyLocalProperty)
* InstanceHolder.InstanceMemberAccess (e.g. MyLocalVar.GetEnumerator or Me.LocalProperty)
* StaticReference (e.g. MyNamespace.NestedNamespace.MyClass(Of Integer).NestedClass(Of String))
* StaticReference.InstanceHolder (e.g. MyNamespace.NestedNamespace.MyClass(Of Integer).NestedClass(Of String).MyStaticPropertyHoldingAnInstance
* StaticReference.InstanceHolder.InstanceMemberAccess (e.g. MyNamespace.NestedNamespace.MyClass(Of Integer).NestedClass(Of String).MyStaticPropertyHoldingAnInstance.MyInstanceProperty.AnotherInstanceProperty)

Any part can contain generics (either as part of a type in a StaticQualifiers or as part of a method call's type parameters in other parts)
If StaticReference and InstanceHolder are both present, StaticReference must end in a type name

### DIFF
--- a/ICSharpCode.CodeConverter/CSharp/DocumentExtensions.cs
+++ b/ICSharpCode.CodeConverter/CSharp/DocumentExtensions.cs
@@ -1,8 +1,11 @@
 ﻿using System;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using ICSharpCode.CodeConverter.Util;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Operations;
 using Microsoft.CodeAnalysis.Simplification;
 using VBasic = Microsoft.CodeAnalysis.VisualBasic;
 using VBSyntax = Microsoft.CodeAnalysis.VisualBasic.Syntax;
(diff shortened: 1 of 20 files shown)


GrahamTheCoder:
@mrmonday Just tagging you here so you don't end up doing anything with the same aim

mrmonday:
Thanks for the heads up!

If you ping me again when you're happy with it I can run your branch against the real world project I've been using and let you know if there are any regressions, if you'd like.

GrahamTheCoder:
Yeah thanks that'd definitely be useful since it's a big change which could easily break all sorts of edge cases, I'll ping you when the PR is ready

GrahamTheCoder:
So I thought I was just about there...but then...

One of the main test cases I wanted to get working was the ReferencingFormThroughStatic part of the solution level tests. Unfortunately the simplification service implementation seems unable to handle this case and on top of that turns into into something that doesn't compile - it prefixes MyBase - doh!
https://github.com/dotnet/roslyn/blob/master/src/Workspaces/VisualBasic/Portable/Simplification/VisualBasicSimplificationService.Expander.vb#L657

What I can't figure out, is how to trigger the expander service within visual studio. I thought that all its refactorings (e.g. method extraction) worked by expanding, refactoring then reducing, but I can't repro the bug there. It's possible that the bug is present but that the reducer manages to reverse the damage done and hence it's never been noticed. This won't work in our case since C# requires more qualification. I'll dig a bit more then:
* boil this down to a minimal repro to file a bug
* workaround it by manually partially qualifying names just enough myself to avoid this bug

GrahamTheCoder:
Introduced issues:
* [ ] MyBase instead of forms within invocation
* [ ] Double qualification (related to ambiguous references perhaps)
* [ ] Complicated generics:
  * `String.Join("|", bs.OrderBy(Function(b) b))` (bs is a `HashSet(Of Double)`)
  * `string.Join<double>("|", Enumerable.OrderBy<double>((IEnumerable<double>)bs, (Func<double, double>)b => b))`
* [ ] Empty "global::" generic for anonymous types

Not fixed (probably need to remove some of the code from QualifyNode):
* [ ] Using statement should not be added Style should have been fully
* [ ] Incorrect static overqualification

GrahamTheCoder:
I've disabled the vast majority of the new functionality, just getting this merged as a foundation for future work (to avoid conflicts) while I troubleshoot using the simplification service. When I've got to the bottom of these issues I'll open a new PR for some proper testing.

Issue #248, opened 2019-02-28
C# -> VB: Namespace not converted correctly


jdmichel:
VB namespaces are appended to the namespace from the project settings.
So, if the project settings have namespace A.B.C, and the original C# code had:
```
namespace A.B.C.D {
     class E {
     }
}
```
Then the correct VB translation is:
```
Namespace D 
     Class E 
     End Class
End Namespace
```
The current conversion with `Namespace A.B.C.D` results in a full class name of A.B.C.A.B.C.D.E.

Converter version 6.4

GrahamTheCoder:
Thanks for the report. I'm very happy to accept a PR, but if not, I will try to get around to it at some point 😄

GrahamTheCoder:
I believe https://github.com/icsharpcode/CodeConverter/pull/450 fixed this

PR #450, opened 2019-11-29
Patch RootNamespace and LangVersion tags, allow customize CSToVBConversion options

Alexilia:
- RootNamespace and LangVersion project's tags are generated accordingly with ILanguageConversion options
- make customizable compiler and parse options to converted VB project in the CSToVBConversion

### Problem
1. A content of the LangVersion tag is different in C# and VB project files.
RootNamespace project's tag have different meaning in C# and VB

2. Using GlobalImports in VisualBasicCompiler.VisualBasicCompilationOptions leads to compilation error in output code, because Simplifier reduces qualified type's names to simple names.

### Solution
1. Patch output vbproj with these properties from CSToVBConversion.VBParseOptions and CSToVBConversion.RootNamespace

2. Note that VisualBasicCompilationOptions.GlobalImports is not a default Imports to any code file, so I decided to not add its as done in VB -> C# converter's part. But making the VisualBasicCompilationOptions.GlobalImports collection empty in VisualBasicCompiler, brakes many tests (especially in VB -> C#), so I pass VisualBasicCompilationOptions with empty GlobalImports to some tests.
CodeConverter.Tests.VB.MemberTests.ParameterWithNamespaceTest describes the problem.
Others tests in CodeConverter.Tests.VB.MemberTests are fixed, because "String" and "DateTime" without namespace "System" is not a keyword in C# so test's input was wrong


Alexilia:
Tests with projects never run correctly.

GrahamTheCoder:
Hi @Alexilia. Thanks for the PR, I'll have a proper look through it soon.

When test outputs need updating, see here how to do it automatically:
https://github.com/icsharpcode/CodeConverter/blob/master/Tests/TestConstants.cs#L17

The project and solution tests fail maybe about 5% of the time due to a bug in the MSBuildWorkspace. When it does this, the output error will very clearly say that's why it's failing.

However, the reason it's failing here is because the test needs updating to take into account the changes you made. You can tell by looking at the test failure message: https://icsharpcode.visualstudio.com/icsharpcode-pipelines/_build/results?buildId=584&view=ms.vss-test-web.build-test-results-tab&runId=1002986&resultId=100342&paneView=debug

Again, using the mechanism above will set a new expected output which you can commit (if it looks correct to you).

If there are any improvements I can make to  [CONTRIBUTING.md](https://github.com/icsharpcode/CodeConverter/blob/master/.github/CONTRIBUTING.md) to make things more obvious let me know!

Alexilia:
No, that is because the MSBuildLocator.RegisterDefaults() method register earlier version of VS.

GrahamTheCoder:
Ah I see, you mean the tests were failing locally - thanks for finding (and fixing!) that

GrahamTheCoder:
I merged the two unrelated commits in https://github.com/icsharpcode/CodeConverter/commit/fb3ffcbfe78605c770d2a165f6aa3ff3c81a6936
Still reviewing the main change.

GrahamTheCoder:
Fixes #248

GrahamTheCoder:
The changes look good. I'm just going to split the ilanguageconversion implementations because they're getting a bit big (already were before this pr, but this has reminded me)
Did you intend to change the VB to Cs conversion? I'll probably change that bit back, but could you explain your understanding of the differences between root name space in VB and Cs? I want to check I'm not missing something

GrahamTheCoder:
Rebased and merged in 2562548c240d3f46950e74f8e9df2f7118659e4b, thanks!

I'd strongly recommend that you (re)base any new branches on origin/master rather than your parallel master branch.

Alexilia:
Thanks for reviewing and merging the code. I have no immediate plans to change VB to Cs conversion. In VB, root namespace are inserted before any declared namespace, while in Cs the default namespace, that defining with the same as in VB RootNamespace tag, doesn't inserted, and affects resources' embedding paths.

GrahamTheCoder:
Great, thanks for that. I hadn't thought about resources. Your original change did affect the conversion in both directions (as you can tell from the test files that got changed by recharacterization), but I've undone the VB to CS change now which I think is correct. Thanks!

Issue #455, opened 2019-12-09
C# -> VB: Inner namespaces are moving to root Imports clause as is


Alexilia:
### Input code
```csharp
namespace System {
    using Collections;

}
```

### Erroneous output
```vbnet
Imports Collections

Namespace System

End Namespace
```

### Expected output
```vbnet
Imports System.Collections

Namespace System

End Namespace
```

### Details

GrahamTheCoder:
Fixed in #456

Issue #454, opened 2019-12-09
C# -> VB: Generic interfaces has wrong Implements statement without type parameters


Alexilia:
### Input code
```csharp
public class TestClass : ITestInterface<string> {
    public void Method(List<string> list) {
}
```

### Erroneous output
```vbnet
Public Class TestClass
    Implements ITestInterface(Of String)
    Public Sub Method(ByVal list As List(Of String)) Implements ITestInterface.Method
    End Sub
End Class
```

### Expected output
```vbnet
Public Class TestClass
    Implements ITestInterface(Of String)
    Public Sub Method(ByVal list As List(Of String)) Implements ITestInterface(Of String).Method
    End Sub
End Class
```

### Details

GrahamTheCoder:
Fixed in #456 (If you write "fixes #456" in the PR it'll autoclose the issue when merged)

Issue #451, opened 2019-12-04
C# -> VB: generic static classes are converted to modules


Alexilia:
### Input code
```csharp
public static class TestClass<T> where T : Class1, new() {
        static Task task;
        static TestClass() {
        }
        public static Task Method() {
            return task;
        }
    }
```

### Erroneous output
```vbnet
Public Module TestClass(Of T As {Class1, New})
    Private task As Task

    Sub New()
    End Sub

    Public Function Method() As Task
        Return task
    End Function
End Class
```

### Expected output
```vbnet
Public NotInheritable Class TestClass(Of T As {Class1, New})
    Private Shared task As Task

    Shared Sub New()
    End Sub

    Public Shared Function Method() As Task
        Return task
    End Function
End Class
```

### Details
there is compilation error because modules can't be generic

GrahamTheCoder:
We should probably convert it to a Class and make all the members Shared to match the behaviour in this case.
We shouldn't do that for all Modules though since it'd break extension methods. Fortunately extension methods aren't allowed in generic classes.

GrahamTheCoder:
Fixed in #456

PR #456, opened 2019-12-09
C# -> VB: don't convert generic static to module, fix generic interface implements statement, concatenate inner namespaces with the parent

Alexilia:
https://github.com/icsharpcode/CodeConverter/issues/451
https://github.com/icsharpcode/CodeConverter/issues/454
https://github.com/icsharpcode/CodeConverter/issues/455

### Problem
1. [generic static classes are converted into modules, which can't be generic](https://github.com/icsharpcode/CodeConverter/issues/451)
2.  [Generic interfaces has wrong Implements statement without type parameters](https://github.com/icsharpcode/CodeConverter/issues/454)
3. [Inner namespaces are moving to root Imports clause as is, but in some cases its should merged with parent namespace](https://github.com/icsharpcode/CodeConverter/issues/455)

### DIFF
--- a/ICSharpCode.CodeConverter/VB/CommonConversions.cs
+++ b/ICSharpCode.CodeConverter/VB/CommonConversions.cs
@@ -421,7 +421,7 @@ namespace ICSharpCode.CodeConverter.VB
             switch (context) {
                 case TokenContext.InterfaceOrModule:
                 case TokenContext.MemberInModule:
-                    return m.IsKind(CSSyntaxKind.StaticKeyword);
+                    return m.IsKind(CSSyntaxKind.StaticKeyword) || m.IsKind(CSSyntaxKind.PublicKeyword);
             }
             return false;
         }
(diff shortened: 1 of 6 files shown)


Alexilia:
I had to download and push these files manually :(

GrahamTheCoder:
OK, I've done the rebase on the above branch and merged, thanks!

To tidy things up for your current work in progress branch "work" could you make sure you've committed any local changes (or reset if you don't want them) then run:
```sh
git fetch --all
git checkout work
git rebase --onto origin/master origin/badbase work
git push --force-with-lease
```
In GitExtensions graph view, the change looks like this, before:
image

After:
image

Once done, if you always just start your branch from origin/master it will keep the list of changes in the PR nice and simple (like it now looks above since I did essentially the same to this PR).

Let me know if you're struggling with any bit of it or what tools you're using and I'll try to give more specific help.

PR #460, opened 2019-12-11
AngularJS 1.5 -> Vue 2.6.10

christophwille:
The Web site was built using a super-old version of Angular - even AngularJS. This is a modernization attempt for two reasons: smaller JS download (150KB down to 30KB), smaller fw for easier use in this super-simple "app".


GrahamTheCoder:
Note: I later added this fixup to restore ctrl+enter functionality which previously relied on jQuery: https://github.com/icsharpcode/CodeConverter/commit/ea421d7ff602a6867d027bdce6b5c6a95dd13a39

Thanks @christophwille!

PR #462, opened 2019-12-12
C# -> VB: Improve CoalesceExpression conversion

Alexilia:
### Problem
ugly CSharpImpl.__Assign() in converted code

### Solution
replace

(x = AssignX())

in expression:

x ?? (x = AssignX())

with:

Function
  x = AssignX()
  Return x
EndFunction


GrahamTheCoder:
I've reopen this PR with some changes I think it needs in order to work, but unfortunately the "Simplifier" stage that runs at the end is breaking it by parentheses required for the program to function correctly. I've been having a few issues with the Simplifier recently - I'll hopefully get around to opening a bug against the roslyn repo - it's possible there's a way around these issues.

PR #464, opened 2019-12-14
Alexilia coalescing assignment

GrahamTheCoder:
Reopening after unmerging #462
I've reopen this PR with some changes I think it needs in order to work, but unfortunately the "Simplifier" stage that runs at the end is breaking it by parentheses required for the program to function correctly. I've been having a few issues with the Simplifier recently - I'll hopefully get around to opening a bug against the roslyn repo - it's possible there's a way around these issues.

Original PR details follow

### Problem
ugly CSharpImpl.__Assign() in converted code

### Solution
replace

(x = AssignX())

in expression:

x ?? (x = AssignX())

with:

Function
  x = AssignX()
  Return x
EndFunction


GrahamTheCoder:
I have started producing a repro of this issue here for submission to Microsoft. I will try to find the time to write up the issue soon:
https://github.com/icsharpcode/CodeConverter/compare/0b596ae0638cba06c728432ae644ac4faef07be5...roslyn-simplifier-issues

Alexilia:
Great job, thanks. I didn't check it in runtime, so got unworking code.
May be we should convert ?? to if else block, and don't wait or work around the simplifier's issue?

GrahamTheCoder:
I think converting to an if-else would be a bit of effort, there could be many null coalesce and assignment operators in a single expression. Though obviously it's possible. Working around the simplifier issue is the best direction. If you have time to implement the workaround to skip annotating invocations with empty argument lists that'd be great, otherwise I'll get around to it in the next week or so hopefully (I'll assign myself to this PR if I'm working on it).
This is the bit you'll need to exclude things from: https://github.com/icsharpcode/CodeConverter/blob/master/ICSharpCode.CodeConverter/CSharp/DocumentExtensions.cs#L27

I'll write up the issue on the roslyn repo now.

Alexilia:
I had prepared the workaround [here](https://github.com/icsharpcode/CodeConverter/pull/467)

GrahamTheCoder:
Great I'll have a final look and get this merged soon

PR #470, opened 2019-12-26
C# -> VB: Remove extra parentheses around CType expression

Alexilia:
### Problem
Parentheses around CType expression lead to compilation error
((Test)o).TestMethod(); ->  (CType(o, Test)).TestMethod()

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * Avoid some case conflicts
 
 ### C# -> VB
-
+* Remove extra parentheses around CType expression
 
 ## [7.4.0] - 2019-12-17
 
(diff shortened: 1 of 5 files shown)


GrahamTheCoder:
Thanks for the xmas **pr**esent! 🎁 

There may well be a wider problem with parenthesization here since VB does like a line starting with a parenthesized expression (from what I can tell). It'd be good to deal with such a problem in a general place. e.g. in VisitStatement might want to unpack any parenthesized expression that turns up so that we don't need to consider it everywhere.

I think the test failure is related to marking all delegates as events which you introduced here: https://github.com/icsharpcode/CodeConverter/pull/442#issuecomment-558114059

So if you're still happy with that just update the test case and let me know if you want to do anything with VisitStatement before merge.

Alexilia:
Merry christmas! Next week I will prepare a fix for redundant event keyword, and if I find more common case to remove parentheses, I will add this too.

GrahamTheCoder:
Thanks again, merry christmas!

Issue #469, opened 2019-12-18
Improve website conversion time


GrahamTheCoder:
Upvote if you want this to happen sooner!

### Summary
With recent major additions, one-file-at-a-time performance is definitely slower. The test time has gone from 59 to 87 seconds since 2019-09-24. Obviously it's a bit up and down but there aren't that many more tests and there's an upwards trend.

While single file performance has decreased, I parallelized at the file-level. So I'm hoping that in the case that takes longest - converting whole projects/solutions - that the overall impact is acceptable for now. The snippet converter seems especially slow even for tiny samples so there may be something bad going on there.

### Cause
One really big cost is GetCompilationAsync within InitializeSourceAsync.

I expected the performance to decrease when adding things like extra separate passes added for things like normalizing inputs: https://github.com/icsharpcode/CodeConverter/pull/386
And when adding conversion of the VB compiler-inlined classes: https://github.com/icsharpcode/CodeConverter/pull/387
So some of the slowdown is just due to doing significantly more work in order to get a more accurate conversion.
However, there are certainly other changes which could potentially be optimized, for example in CS->VB: PR #456 introduces a potentially slightly slow operation on a "warm" path: https://github.com/icsharpcode/CodeConverter/pull/456/files#diff-709371109bde62ae9a9082fda70f3a94R192

There are some more big changes coming in a general drive to get VB->CS winforms conversion pretty solid. For example once I've flipped this switch it will go even slower: https://github.com/icsharpcode/CodeConverter/blob/master/ICSharpCode.CodeConverter/CSharp/DocumentExtensions.cs#L45
However, it should allow deleting some other bits of code on hot paths, which will speed things up.

### Solution

Do some profiling and optimize as appropriate.

GrahamTheCoder:
I'm now happy that project level specific   code performs well enough. The next step would be to focus on reducing time to convert for a single typical small snippet that'd be used in the Web converter

GrahamTheCoder:
For some reason the live site is significantly (5x) slower for the trivial sample conversion than running the website locally

The local site is slow only for the first request, then fast (with browser cache disabled)

christophwille:
I played around a bit with (unscientifically) Task Manager. Memory for the worker process is negligible, however, the CPU sometimes goes up considerably for the process (up to 30-40%). That's when the slowdown is noticeable (could parallel conversion tax the CPU that much?).

GrahamTheCoder:
[...]

PR #471, opened 2019-12-29
Enable expand phase

### LINKED ISSUE #473: VB -> C#: Duplicated property names (for prop in base class)
### Erroneous output
UseWaitCursor.UseWaitCursor 

### Expected output
Application.UseWaitCursor

GrahamTheCoder:
Discussed in #449

### VB -> CS
* [x] Fix winforms qualification issue (should be some valid subpart of `Global.WindowsAppVb.My.MyProject.Forms`)
* [x] Not part of tree bug https://github.com/icsharpcode/CodeConverter/pull/471#issuecomment-570410565
* [x] Binary expression overcompensation https://github.com/icsharpcode/CodeConverter/pull/471#issuecomment-570411599

### CS -> VB
* [x] Fix overqualification with global
* [x] Re-enable case conflict test that this fixes
* [x] Workaround hard crash - fatal CLR/codeanalysis bug for invalid tree https://github.com/icsharpcode/CodeConverter/pull/471#issuecomment-570525721
### Notes
* Now errs on the side of type name rather than var, this could go wrong with anonymous types perhaps. I'd like to move this back towards using var more, but only when we're *sure* it's sensible.
* In VB->CS, casts appear in more of the required places and fewer redundant places
* Fixes #473


GrahamTheCoder:
@mrmonday This switches on the "Expand" phase previously mentioned. If you have any real test cases to throw at it before merge that'd be great. The highest priority to watch out for are new syntax errors introduced, so it's probably worth converting with current master, committing the result to git, then diffing the result produced by this version to make it easy to spot differences.

If anything is consistently annoyingly overly verbose now that's also worth knowing though may be worth accepting for the benefits. I'm aware this really doesn't help performance but will deal with that separately in #469

I had to workaround about 5 roslyn issues to get this usable, so there could easily be more lurking. Given the issues with this part it's unlikely I'll add any other node types to the list of expanded nodes unless we have specific reason to do so.

mrmonday:
I'm travelling today, but I'll run some conversations in the next couple of days and let you know how it goes :)

GrahamTheCoder:
Just found this bug, likely due to this PR since I visited the Instance Syntax node from the operation:
```
#error Cannot convert MultiLineIfBlockSyntax - see comment for details
            /* Cannot convert MultiLineIfBlockSyntax, System.ArgumentException: Node is not within syntax tree
               at Microsoft.CodeAnalysis.VisualBasic.VBSemanticModel.CheckSyntaxNode(VisualBasicSyntaxNode node)
               at Microsoft.CodeAnalysis.VisualBasic.VBSemanticModel.GetSymbolInfo(ExpressionSyntax expression, CancellationToken cancellationToken)
               at Microsoft.CodeAnalysis.VisualBasic.VBSemanticModel.GetSymbolInfoForNode(SyntaxNode node, CancellationToken cancellationToken)
               at Microsoft.CodeAnalysis.VisualBasic.VBSemanticModel.GetSymbolInfoCore(SyntaxNode node, CancellationToken cancellationToken)
               at ICSharpCode.CodeConverter.CSharp.ExpressionNodeVisitor.<VisitMemberAccessExpression>d__43.MoveNext()
```

Cut down repro:
```vbnet
If identifierNameForAForm IsNot Nothing Then
    If identifierNameForAForm.AProperty IsNot Nothing Then
        identifierNameForAForm.AProperty.LogError(identifierNameForAForm.AProperty.AnotherProperty, "Error: " & ex.Message)
    End If
    My.Computer.Clipboard.SetText("")
End If
```

GrahamTheCoder:
An interesting change:
```vbnet
If anEnumTypedParameter <> Nothing Then
    Console.WriteLine("")
End If
```
used to be
```csharp
if (anEnumTypedParameter != default(int))
...
```

now is
```csharp
if ((int)anEnumTypedParameter != default(TheEnumType))
```

GrahamTheCoder:
CS->VB: 
Need to dig into repro's parallel stacks to narrow it down and see if it's specific to this branch

```csharp
Managed Debugging Assistant 'FatalExecutionEngineError' 
  Message=Managed Debugging Assistant 'FatalExecutionEngineError' : 'The runtime has encountered a fatal error. The address of the error was at 0x36965e72, on thread 0x5244. The error code is 0x80131623. This error may be a bug in the CLR or in the unsafe or non-verifiable portions of user code. Common sources of this bug include user marshaling errors for COM-interop or PInvoke, which may corrupt the stack.'

System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.CodeAnalysis.VisualBasic.Binder.OptimizeSelectStatement(BoundRValuePlaceholder selectExpression, ArrayBuilder`1 caseBlockBuilder, Boolean& generateSwitchTable, DiagnosticBag diagnostics)
   at Microsoft.CodeAnalysis.VisualBasic.Binder.BindCaseBlocks(SyntaxList`1 caseBlocks, BoundRValuePlaceholder selectExpression, Boolean convertCaseElements, Boolean& recommendSwitchTable, DiagnosticBag diagnostics)
   at Microsoft.CodeAnalysis.VisualBasic.Binder.BindSelectBlock(SelectBlockSyntax node, DiagnosticBag diagnostics)
   at Microsoft.CodeAnalysis.VisualBasic.Binder.BindStatement(StatementSyntax node, DiagnosticBag diagnostics)
   at Microsoft.CodeAnalysis.VisualBasic.Binder.BindBlock(SyntaxNode syntax, SyntaxList`1 stmtList, DiagnosticBag diagnostics, Binder stmtListBinder)
   at Microsoft.CodeAnalysis.VisualBasic.Binder.BindBlock(SyntaxNode syntax, SyntaxList`1 stmtList, DiagnosticBag diagnostics)
   at Microsoft.CodeAnalysis.VisualBasic.Binder.BindMethodBlock(MethodBlockBaseSyntax methodBlock, DiagnosticBag diagnostics)
   at Microsoft.CodeAnalysis.VisualBasic.Binder.BindStatement(StatementSyntax node, DiagnosticBag diagnostics)
   at Microsoft.CodeAnalysis.VisualBasic.Symbols.SourceMethodSymbol.GetBoundMethodBody(TypeCompilationState compilationState, DiagnosticBag diagnostics, Binder& methodBodyBinder)
   at Microsoft.CodeAnalysis.VisualBasic.Symbols.SourceMemberMethodSymbol.GetBoundMethodBody(TypeCompilationState compilationState, DiagnosticBag diagnostics, Binder& methodBodyBinder)
   at Microsoft.CodeAnalysis.VisualBasic.MethodCompiler.BindAndAnalyzeMethodBody(MethodSymbol method, TypeCompilationState compilationState, DiagnosticBag diagnostics, Binder containingTypeBinder, MethodSymbol& referencedConstructor, Boolean& injectDefaultConstructorCall, Binder& methodBodyBinder)
   at Microsoft.CodeAnalysis.VisualBasic.MethodCompiler.CompileMethod(MethodSymbol method, Int32 methodOrdinal, Int32& withEventPropertyIdDispenser, Int32& delegateRelaxationIdDispenser, Predicate`1 filter, TypeCompilationState compilationState, ProcessedFieldOrPropertyInitializers processedInitializers, Binder containingTypeBinder, SynthesizedSubmissionFields previousSubmissionFields, MethodSymbol& referencedConstructor)
   at Microsoft.CodeAnalysis.VisualBasic.MethodCompiler.CompileNamedType(NamedTypeSymbol containingType, Predicate`1 filter)
   at Microsoft.CodeAnalysis.VisualBasic.MethodCompiler._Closure$__30-0._Lambda$__0()
```

mrmonday:
It seems my Windows box has decided to not work any more, so I will unfortunately be unable to test until next week.

I'll give it a go on Monday.

GrahamTheCoder:
Oh dear, no worries I've tested it on a couple of things and I think it's OK. I'll merge  since it seems no worse than before and several things are better. But it'd still be great to run it against some real projects - just raise any bugs you see as new issues and if you find something really bad we can always disable the new behaviour for the direction it occurs in.

#### Behaviour change
I've just converted ~40k lines of CS and come across an interesting change. There was an existing bug, which got fixed, but it introduces another similar bug at the same time.

It doesn't add the `Imports` in either case, but now qualifies part of it correctly but adds another incorrectly qualified part.


```csharp
public class QualificationTest
{
    private static void AMethod()
    {
        System.Reflection.Assembly a = System.Reflection.Assembly.GetExecutingAssembly();
    }
}
```

7.4:
```vbnet
Public Class QualificationTest
    Private Shared Sub AMethod()
        Dim a = Assembly.GetExecutingAssembly
    End Sub
End Class
```

This PR:
```vbnet
Public Class QualificationTest
    Private Shared Sub AMethod()
        Dim a As Assembly = System.Reflection.Assembly.GetExecutingAssembly()
    End Sub
End Class
```

mrmonday:
There are definitely regressions from `7.3.0` (last version I used and confirmed working) -> `7.4.0`, and from `7.4.0` -> `master`.

Unfortunately, the code-base I'm using to test has also introduced some more VB-isms which can't be converted.

I'll find an older version of the code, make sure it converts with `7.3.0`, then work my way forwards and file some bugs.

mrmonday:
There seem to be a bunch of issues related to types in nuget packages that didn't happen with older versions - I can't seem narrow any down to a test case though.

mrmonday:
I'm done reporting issues for now - there seems to be one real issue caused by this, and a bunch of issues caused by missing semantic information for nuget packages (which may or may not be related to this).

I found an issue for code that looks something like:

```vbnet
Dim x As Thing(Of Foo) = A.B.C(something)
```

Becomes:

```c#
Thing x = A.B.C(something);
```

Instead of:

```c#
Thing<Foo> x = A.B.C(something);
```

But I couldn't narrow it down. It could be related to the missing semantic info.

GrahamTheCoder:
That repro works for me on 7.4 when the type of Foo is missing. Probably a common case to try to improve anyway though

Saibamen:
Also fixes https://github.com/icsharpcode/CodeConverter/issues/475

PR #472, opened 2019-12-31
C# -> VB: fix wrong event keyword

Alexilia:


GrahamTheCoder:
Thanks, I'll try to take a look some time this weekend

Alexilia:
I try to use another approach to mark delegates with event keyword. I am analyzing is there any events' symbols in containing type, which use the delegate.

GrahamTheCoder:
Great, thanks I'll try to take a look this weekend

GrahamTheCoder:
Implementation-wise this looks good now it's using the semantic model. The only other thing I'd do is try to move the analysis of which delegate fields are assigned to the VisitCompilation unit phase so it only happens once rather than N times.

However, I've done a bit of poking around the docs on VB and [C# events](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/event), and I've thought of a test case which makes me question whether we want to add the event keyword anywhere. Consider the case where the delegate is used for other things:
```csharp
class EventClassWithReset
{
    EventHandler backingField;
    Func<string> createSomething;
    public event EventHandler MyEvent
    {
        add { backingField += value; }
        remove { backingField -= value; }
    }

    public void Reset()
    {
        backingField = null;
    }
}
```

I think it makes most sense to change the "+=" and "-=" onto Delegate.Combine and Delegate.Remove which is [I think what it will compile to in C#](https://stackoverflow.com/a/33964976/1128762) - haven't actually checked.
```vbnet
Class EventClassWithReset
    Private backingField as EventHandler
    Private createSomething As Func(Of string)

    Public Custom Event MyEvent As EventHandler
        AddHandler(ByVal value As EventHandler)
            backingField = [Delegate].Combine(backingField, value)
        End AddHandler
        RemoveHandler(ByVal value As EventHandler)
            backingField = [Delegate].Remove(backingField, value)
        End RemoveHandler
        RaiseEvent(ByVal sender As Object, ByVal e As EventArgs)
            backingField(sender, e)
        End RaiseEvent
    End Event

    Public Sub Reset()
        backingField = Nothing
    End Sub
End Class
```

So it seems C# just has a bit of extra syntactic sugar around delegates which we could expand in VB:
```csharp
        Action x = () => { };
        x = x + x;
```

```vbnet
        Dim x As Action = Sub()
        x = [Delegate].Combine(x, x)
```

Thoughts? (Sorry for not raising this sooner!)

Alexilia:
I have added implementation with Delegate. I think it acceptable. Thanks for advice.

GrahamTheCoder:
Great, that's looking a lot better. Thanks for following up on this!

Before I merged, I just replaced the "Parse" methods with other SyntaxFactory calls. You can see that and other advice in the [contributing doc](https://github.com/icsharpcode/CodeConverter/blob/master/.github/CONTRIBUTING.md#implementation-advice). I've just added these two tools for visualizing source trees: [Syntax Vizualizer](https://marketplace.visualstudio.com/items?itemName=VisualStudioProductTeam.NETCompilerPlatformSDK) and the [sharplab website](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQBsQDdhJiaMkAmIA1AD4CSYADgPYIxQAEAygJ6xxgCwAUAGIIAlnAB2BRgGEU0JgBU4sKTN6NVjAApD0EeI2AQAxgGtBIgOZ9hKcQEEmAUXSiYACQhiUiFWvUBXYCiCBpK+sDRgjI7OjACyrFEiMIx2kU6Jbh5ePGo5yQQEGQSeCAAUAEKsAGoQKIzaKL5wyQ5pru5FiACU3rm5+samFlbiALyMANoAInCeZjpwALoAdBLhwKZwJf0m5pYzBEh1NY3d2b2RYnkF7cU9uQBKnDROhcXlVTVHDU0pCW2ZCFO5z6hh2Q32jDGUxmcDm8GWjzAz0220Ge2sh3qJzuOXsl0RyNeWWB9wggigcD+72qtQpYkQzUYAHlgAArOAGfCMCo0xg/FrOGwIMxQIHAtSo3bDAD8JTpBEQhzgYt6ePEpPJlNadzVqWcdz8ASCLH8jEeFJgJRVIIGUohYwAcjQYAALQY6y7Mfy8XVKKBQIA===). Hope it helps, thanks again!

GrahamTheCoder:
Ah it turns out that the simplifier has a bug for that type of conditional invocation anyway so the parsing is a good workaround for that case for now. Good job!

PR #499, opened 2020-01-15
Nuspec dependencies

GrahamTheCoder:
### Problem
* The core library targets old versions of microsoft.codeanalysis in order to allow compatibility with old versions of visual studio (for both our own vsix, and other vsixs consuming this package).
* The core library also targets netstandard1.3 for this reason.
* By nuget's design, people who don't have such targeting requirements are also getting old dependency versions by default under this scheme unless they use a [nuget.config](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file) with dependency version set to highest.
* We feel a bit bad about propagating ancient versions of things

### Solution
* For people targeting netstandard2.0 or above, give them a later version (that our tests run against)

### Details
* I found the page here useful for checking which properties exist: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target/
* We may want to put a [nuget.config](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file) in our repository with dependency version set to highest to ensure we're testing against the latest consumers would get if they did the same

### To do
* [x] Set the version for netstandard2.0 to be the earliest version that supported netstandard2.0
* [ ] Test the nupkg - should give 2.6.1 if you install in a project below nestandard2, otherwise the 3.0 packages

### DIFF
--- a/ICSharpCode.CodeConverter/ICSharpCode.CodeConverter.csproj
+++ b/ICSharpCode.CodeConverter/ICSharpCode.CodeConverter.csproj
@@ -4,24 +4,17 @@
     <AssemblyName>ICSharpCode.CodeConverter</AssemblyName>
     <RootNamespace>ICSharpCode.CodeConverter</RootNamespace>
     <PackageTargetFallback>portable-net45+win8</PackageTargetFallback>
-    <Company>ICSharpCode</Company>
-    <Description>Convert VB.NET to/from C#
-
-	* Accurate: Full project context (through Roslyn) is used to get the most accurate conversion.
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
Roslynator just declared their final code analysis 2.x release (they were on about 2.3) and jumped to 3.0
https://github.com/JosefPihrt/Roslynator/blob/master/src/CSharp/CSharp.csproj

GrahamTheCoder:
Instead of this I'm going to move to netstandard2.0 with the code converter 8.0 release
In November, when VS2017 (15.9) is out of mainstream support, I'll move the codeanalysis libraries forward to 3.0.

PR #505, opened 2020-01-17
C# -> VB: rename conflicting members inside types

Alexilia:


GrahamTheCoder:
Looks great! I did some renaming and moving stuff around, but the only functionality I tweaked was to try to work better with multi-location symbols like partial classes and to prefer naming local symbols over parameters. I think it'll still have some issues in complex scenarios, but we can deal with them as they come up. Thanks 😄

PR #511, opened 2020-01-24
C# -> VB: Small improvements

Alexilia:


Tests/VB/StatementTests.cs
```diff
@@ -755,6 +755,90 @@ void TestMethod()
...
+    Private Sub TestMethod()
+        Dim summary = 0
+
+        For month As Integer = 1 To 12
+            summary += month
+        Next
+    End Sub
+End Class");
+        }
+        [Fact(Skip = "Should we rename, or add If True End If block?")]
```
GrahamTheCoder: My order of preference would be:
1) Rename - most effort to implement, but we do already have a renaming phase that looks at locals
2) If True - Gets it compiling and keeps the loops isolated, but someone would probably come along and manually rename afterwards
3) Remove the Dim keyword in the second case - Gets it compiling but makes the code a bit harder to maintain

Tests/VB/StatementTests.cs
```diff
@@ -755,6 +755,90 @@ void TestMethod()
     End Sub
 End Class");
         }
+        [Fact(Skip = "month As Integer is reduced by Simplifier, and month threated as Date.Month method")]
```
GrahamTheCoder: I think this is likely  caused by the fact that the converter currently puts a bunch of Imports guesses directly into the file which probably makes the symbol ambiguous in the initial parsing.

GrahamTheCoder:
Yep all looks sensible to me

Tests/VB/StatementTests.cs
Alexilia: It is because of type promotion. There is the DateAndTime module in Microsoft.VisualBasic namespace, and its members are promoted to Microsoft.VisualBasic namespace. Microsoft.VisualBasic namespace  seems to imported implicitly. But the same problem can occur with any module if it's namespace is imported.

GrahamTheCoder: > Microsoft.VisualBasic namespace seems to imported implicitly

Exactly - all of these get put at the top of every file:
https://github.com/icsharpcode/CodeConverter/blob/master/ICSharpCode.CodeConverter/Util/VisualBasicCompiler.cs#L71

PR #513, opened 2020-01-27
C# -> VB: Global qualification problem

Alexilia:
GlobalNameSyntax contains in the last leaf of QualifiedNameSyntax's nodes tree.

Simplier reduces erroneous Global syntax in Imports. So, the issue doesn't covered by the GlobalImportsStatement test. Simplifier fails on some code files, while code are compiling and properly running.


GrahamTheCoder:
Ah yes, sorry I missed that the first time. Would be great to get a test case. If you have a file that reproduces the issue, could you try chopping out half at a time until it's small enough to be used as a test case?

PR #515, opened 2020-01-30
C# -> VB: field events with operators += -= can appear in property events

Alexilia:
Have to return AddHandler RemoveHandler RaiseEvent code generation to custom events using field events

### DIFF
--- a/ICSharpCode.CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
+++ b/ICSharpCode.CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
@@ -275,8 +275,11 @@ namespace ICSharpCode.CodeConverter.VB
         public override SyntaxList<StatementSyntax> VisitForStatement(CSS.ForStatementSyntax node)
         {
             StatementSyntax block;
-            if (!ConvertForToSimpleForNext(node, out block)) {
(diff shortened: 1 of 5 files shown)


ICSharpCode.CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
```diff
@@ -294,6 +294,10 @@ public override SyntaxList<StatementSyntax> VisitForStatement(CSS.ForStatementSy
             }
             return SyntaxFactory.SingletonList(block);
         }
+        private static StatementSyntax ReplaceExitForToWhile(StatementSyntax statement) {
+            var exitSyntaxes = statement.DescendantNodesAndSelf().OfType<ExitStatementSyntax>();
+            return statement.ReplaceNodes(exitSyntaxes, (_, __) => SyntaxFactory.ExitWhileStatement());
```
GrahamTheCoder: Using any of the Replace functions on the conversion result from another Visit method causes issues keeping track of where nodes came from. On PR #512 I'm currently unpicking existing ones of these to try to get comment conversion working C# -> VB.

I've added a commit to replace this with something that also works for nested loops and the continue statement. Feel free to build on that if you can think of other cases.

GrahamTheCoder:
I haven't tried out the syntax, but your tests and implementation look reasonable other than the minor edit I made, thanks!

PR #516, opened 2020-01-31
Merge pull request #516 from Alexilia/master

Alexilia:
### Problem
  If a field declared with an initializer, Simplifier reduce the declaration, and variable's type is omitted.

private int aVariable = 1;

converted to

Private aVariable = 1

  In the case, aVariable's type became Object, which can be a reason of many compile-time and run-time errors.

### Solution
  I am using SimplificationOptions.PreferImplicitTypeInLocalDeclaration option is set to false. But I had to fix about 50 tests.
  As side effect, the ForWithDateTimeVariables test is passing now

### DIFF
--- a/ICSharpCode.CodeConverter/Shared/DocumentExtensions.cs
+++ b/ICSharpCode.CodeConverter/Shared/DocumentExtensions.cs
@@ -8,6 +8,7 @@ using ICSharpCode.CodeConverter.VB;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Options;
 using Microsoft.CodeAnalysis.Simplification;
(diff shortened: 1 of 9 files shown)


ICSharpCode.CodeConverter/Shared/DocumentExtensions.cs
```diff
@@ -113,5 +116,11 @@ private static SyntaxNode WithWarningAnnotation(SyntaxNode node, string warningT
         {
             return node.WithAdditionalAnnotations(new SyntaxAnnotation(AnnotationConstants.ConversionErrorAnnotationKind, warningText));
         }
+        private static OptionSet GetVBOptions(DocumentOptionSet options) {
+            return options.WithChangedOption(SimplificationOptions.PreferImplicitTypeInLocalDeclaration, false);
```
GrahamTheCoder: For some reason I couldn't find the way to change the simplification options when I looked, so this is extra awesome! Now I can go and change a bunch of others that cause issues 😁

PR #532, opened 2020-02-21
Move XHR to Axios

christophwille:
Motivation: easier to read (this.readyState === 4 isn't exactly easy to grok without docs), JSON by default, and Axios also doesn't need a Polyfill.


CodeConverter.Web/wwwroot/js/index.js
```diff
@@ -54,12 +53,28 @@ var app = new Vue({
                         output.select();
                     });
                 }
-            };
-            xhttp.open("POST", "/api/converter/", true);
-            xhttp.setRequestHeader("Content-Type", "application/json");
+            })
+            .catch(function (error) {
+                capturedObj.converterCallInflight = false;
```
GrahamTheCoder: Good that we're catching errors now. We could have done so with `xhttp.onError` but never did.
Now that we are, it seems like a good opportunity to set `capturedObj.errorMessageOnResponse` to tell the user what's happened. Then just console.log any extra details not already shown in the UI.

In the XMLHttpRequest world we could have used onError for this.

CodeConverter.Web/wwwroot/js/index.js
```diff
@@ -32,20 +32,19 @@ var app = new Vue({
...
+                code: this.inputCode,
+                requestedConversion: this.requestedConversion
+            })
+            .then(function (response) {
+                capturedObj.converterCallInflight = false;
+                if (200 === response.status) {
+                    capturedObj.convertedCode = response.data.convertedCode;
 
-            xhttp.onreadystatechange = function () {
-                if (this.readyState === 4) {
```
GrahamTheCoder: If the aim was to make this simpler to read, it might have been a bit more lightweight to use the onLoad method which only fires in readystate 4.
Or you could just extract the 4 as a nicely named constant, or set it to `XMLHttpRequest.DONE ||  XMLHttpRequest.READYSTATE_COMPLETE` (the latter part if for IE10 compatibility).

Given the simplicity of the site, I think we're unnecessarily introducing more possible failure states by adding dependencies, especially in terms of security vulnerabilities. That said, it looks like axios is the best choice of the ones available.

christophwille:
I totally agree that any framework is overkill for the given use case. Original motivation (AngularJS version) was to make a [for us non-JS devs] reasonably understandable and "familiar-looking" codebase that wouldn't create too much of a cognitive load when trying to understand what is going on in there.

PR #533, opened 2020-02-25
C# -> VB: various fixes

Alexilia:


ICSharpCode.CodeConverter/VB/CaseConflictResolver.cs
```diff
@@ -107,8 +114,8 @@ private static async Task<Project> PerformRenames(Project project, IReadOnlyColl
                 project = solution.GetProject(project.Id);
                 var compilation = await project.GetCompilationAsync();
                 ISymbol currentDeclaration = SymbolFinder.FindSimilarSymbols(originalSymbol, compilation).FirstOrDefault();
-                if (currentDeclaration == null) break; //Must have already renamed this symbol for a different reason
-
+                if (currentDeclaration == null)
+                    continue; //Must have already renamed this symbol for a different reason
```
GrahamTheCoder: Whoops, good spot!

Tests/VB/ExpressionTests.cs
```diff
@@ -880,6 +880,61 @@ End Sub
...
+        If s1 = j Then DoSomething()
+        If s1 = s2 Then DoSomething()
+        If s1 Is object2 Then DoSomething()
+        If object1 = j Then DoSomething()
+        If object1 Is s2 Then DoSomething()
+        If object1 Is object2 Then DoSomething()
+    End Sub
+
+    Public Sub DoSomething()
+    End Sub
```
GrahamTheCoder: This looks like a good candidate for a self verifying test

ICSharpCode.CodeConverter/VB/NodesVisitor.cs
```diff
@@ -911,13 +911,13 @@ public override VisualBasicSyntaxNode VisitParenthesizedExpression(CSS.Parenthes
...
-                    SyntaxFactory.Token(VBUtil.GetExpressionOperatorTokenKind(kind)), _commonConversions.Literal(1)
-                );
-            }
+            //if (IsReturnValueDiscarded(node)) {
+            //    return SyntaxFactory.AssignmentStatement(
+            //        kind,
+            //        (ExpressionSyntax)node.Operand.Accept(TriviaConvertingVisitor),
+            //        SyntaxFactory.Token(VBUtil.GetExpressionOperatorTokenKind(kind)), _commonConversions.Literal(1)
+            //    );
+            //}
```
GrahamTheCoder: If this bit needs to go, just delete it. We can always get it back from source control if we figure out what on earth it was trying to achieve at some point in the future (it's not obvious to me) 😄

GrahamTheCoder:
Sorry I took so long to get around to this. Looks really good, just a couple of minor suggestions.

Tests/VB/ExpressionTests.cs
GrahamTheCoder: I'll add one in now

GrahamTheCoder: I've pushed the tests. Without your change on NodesVisitor.cs:1208 they pass, but with that change they fail.
Therefore I'n not convinced this is the right conversion. Could you take a look and either explain what problem you're fixing, or revert the change if it's wrong? Thanks!

Alexilia: **Is** operator compares two objects [by reference](https://docs.microsoft.com/en-gb/dotnet/visual-basic/language-reference/operators/is-operator). It does not perform value comparison.
**=** operator with two strings is compiled to IL using [Microsoft.VisualBasic.Strings.Equals](https://anthonydgreen.net/2019/02/12/exhausting-list-of-differences-between-vb-net-c/#63). So the comparison could be culture-aware. In VB null strings and empty strings are considered equal, and the way to compare a string with null is to use **Is Nothing**. In simple cases it can't be noticed because compiler use String.Intern.

GrahamTheCoder: Yep so I think we're agreed the runtime equivalent of CSharp "==" in VB is "Is".
If we're not, could you add a test case to the self verifying tests to show where they differ (to pass, the self verifying test must compile and pass in CSharp *and* after conversion to visual basic)

In terms of priorities for conversion, runtime equivalence is priority 1, idiomatic code in the target language is priority 3:
https://github.com/icsharpcode/CodeConverter/blob/master/.github/CONTRIBUTING.md#deciding-what-the-output-should-be
I. E. Code that works the same is key

Note: In CSharp you would use the Equals method to do invariant culture or case insensitive comparison.

Alexilia: Nop, in CSharp **==** is compiled to IL to **[System.Runtime]System.String::op_Equality**. It is more equivalent of VB **=**. These simple test passes in CS and not in VB with **Is**:
```
[Fact]
    public void NonInternStringsEqualsOperator() {
        string a = "10";
        string b = 10.ToString();
        Assert.True(a == b);
    }
```
```
Public Sub NonInternStringsEqualsOperator()
        Dim a As String = "10"
        Dim b As String = 10.ToString()
        Assert.[True](a Is b)
End Sub
```

but passes with **=**
```
Public Sub NonInternStringsEqualsOperator()
        Dim a As String = "10"
        Dim b As String = 10.ToString()
        Assert.[True](a = b)
End Sub
```

 Empty string and null string in VB are the same because **Nothing** in VB is not **null** in CSharp, it's default value, for example 0 = Nothing for integers, so in any comparison null and empty string should be the same. The runtime functions also treat null like empty, so it should not be a problem.

Tests/TestData/SelfVerifyingTests/CSToVB/StringEqualityTests.cs
```diff
@@ -10,10 +10,17 @@ public class StringEqualityTests
     private char[] emptyCharArray = new char[0];
 
     [Fact]
-    public void NullEmptyStringEqualsOperator() => Assert.False(nullString == emptyString);
+    public void NonInternStringsEqualsOperator() {
+        string a = "10";
+        string b = 10.ToString(); //strings created in runtime are not interned
+        Assert.True(a == b);
+    }
```
GrahamTheCoder: Yep, I like this test, it shows that op_Equality is overridden and distinguishes from reference equality (which neither VB or C# does, but someone might guess from "==")

Tests/VB/ExpressionTests.cs
GrahamTheCoder: Ah, my apologies for misunderstanding. I understand now - I had assumed "Is" called op_Equality, but in fact it calls reference equals. This means the implementation in master is flawed as your new interning related test (presumably) highlights.

But my previous tests you removed/changed show the implementation on this PR is also flawed. You'll need to restore those tests to find the correct solution to this problem. I think the solution is to call `Equals(str1, str2)` when both sides are known to be strings.

We should also ensure there's a test case for object == string that preserves the unintended reference equality comparison like this:
https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUBgBjwwGYACAbzzJrI3UutuYHsAjAKwFMBjYMgB5kAvGUwEAdABUWAZWBgAllADmACgCUAbibMaqcWQCeIsgCJxZnbj20DATjVDhoo9t1kAvnk9A

Saibamen:
Fixed in release 8.0.0

PR #538, opened 2020-03-06
Enable code coverage

GrahamTheCoder:
The idea is to be able to use this as a meaningful metric to show whether the real terms code coverage is improving. This means we should exclude bits which are unused / tested elsewhere

* [x] Should probably exclude the "FromRoslyn" namespace since lots of those methods are unused and are tested within the Roslyn repo
* [x] Remove a bunch of unused code


GrahamTheCoder:
This should be much more useful once https://github.com/microsoft/azure-pipelines-tasks/issues/11677 is released

PR #566, opened 2020-04-26
Byref

### LINKED ISSUE #310: VB -> C#: Conditional method calls with ByRef parameters
Via #307.

### Input code
```vbnet
If False AndAlso MethodWithByRef(MethodWithSideEffects()) Then
    DoSomething()
End If
```

### Erroneous output
```c#
var arg = MethodWithSideEffects();
if (false && MethodWithByRef(ref arg))
{
    DoSomething();
}
```

### Expected output

Unsure. Maybe something like:

```c#
Func<bool> arg = () => { var arg = MethodWithSideEffects(); MethodWithByRef(ref arg); };
if (false && arg())
{
    DoSomething();
}
```

### LINKED ISSUE #324: Logic error when converting ByRef parameters
Originally reported on [a wordpress comment](https://grahamthecoder.wordpress.com/about/#comment-89).

### Input code
```vbnet
Public Class MyTestClass

	Private Property Prop As Integer
	Private Sub TakesRef(ByRef vrbTst As Integer)
            vrbTst = vrbTst + 1
	End Sub

	Private Sub UsesRef(someInt As Integer)
		TakesRef(Prop)
		someInt += 1
	End Sub
End Class
```

### Erroneous output

```csharp
public class MyTestClass
{
    private int Prop { get; set; }
    private void TakesRef(ref int vrbTst)
    {
        vrbTst = vrbTst + 1;
    }

    private void UsesRef(int someInt)
    {
        var argvrbTst = Prop;
        TakesRef(ref argvrbTst);
        someInt += 1;
    }
}
```

### Expected output
```csharp
public class MyTestClass
{
    private int Prop { get; set; }
    private void TakesRef(ref int vrbTst)
    {
        vrbTst = vrbTst + 1;
    }

    private void UsesRef(int someInt)
    {
        var argvrbTst = Prop;
        TakesRef(ref argvrbTst);
        Prop = argvrbTst; // Need to set value of property afterwards to maintain logic
        someInt += 1;
    }
}
```

### Details
Should also:
* Create self-verifying tests to check cases where an implicit type conversion happens
* Check what happens when semantic info is incomplete (that it "errs" in the sensible direction)

See https://github.com/icsharpcode/CodeConverter/blob/c70e6aa421e5c93bcb2fe19c3d428fcad82a0f01/ICSharpCode.CodeConverter/CSharp/NodesVisitor.cs#L1460-L1469

If this solution is implemented, we should also bear in mind #310

### LINKED ISSUE #567: VB -> C#: Unnecessary arg variable and stray ==
### Input code
```vbnet
Public Class VisualBasicClass
    Dim arr(,) As String

    Sub DoSomething(ByRef str As String)
        str = "test"
    End Sub

    Sub Main()
        DoSomething(arr(2, 2))
        Debug.Assert(arr(2, 2) == "test")
    End Sub

End Class
```

### Erroneous output
```C#
using System.Diagnostics;

public partial class VisualBasicClass
{
    private string[,] arr;

    public void DoSomething(ref string str)
    {
        str = "test";
    }

    public void Main()
    {
        string argstr = arr[2, 2];
        DoSomething(ref argstr);
        Debug.Assert(arr[2, 2] ==  == "test");
    }
}
```

### Expected output
```C#
using System.Diagnostics;

public partial class VisualBasicClass
{
    private string[,] arr;

    public void DoSomething(ref string str)
    {
        str = "test";
    }

    public void Main()
    {
        DoSomething(ref arr[2, 2]);
        Debug.Assert(arr[2, 2] == "test");
    }
}
```

### Details
* Product in use: Web converter
* Version in use: Version 8.1.0.0

There are a few issues here:

- [ ] Instance variable not updated (covered in #324)
- [x] Additional `==` in the assert
- [ ] Unnecessary variable for the parameter - it's only needed for properties, not fields

GrahamTheCoder:
Fixes #324 
Fixes #310
Fixes #567


mrmonday:
#567 has another case you might like to cover in this PR.

CodeConverter/CSharp/ExpressionNodeVisitor.cs
```diff
@@ -1287,23 +1418,68 @@ private IEnumerable<ArgumentSyntax> GetAdditionalRequiredArgs(ISymbol invocation
...
+
+            return GetRefConversion(expression);
 
-            var symbolInfo = GetSymbolInfoInDocument<ISymbol>(node.Expression);
-            bool isProperty = symbolInfo != null && symbolInfo.IsKind(SymbolKind.Property);
-            bool isUsing = symbolInfo?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax()?.Parent?.Parent?.IsKind(VBasic.SyntaxKind.UsingStatement) == true;
+            RefConversion GetRefConversion(VBSyntax.ExpressionSyntax expression)
+            {
+                var symbolInfo = GetSymbolInfoInDocument<ISymbol>(expression);
+                if (symbolInfo.IsKind(SymbolKind.Property)) return RefConversion.PreAndPostAssignment;
```
GrahamTheCoder, line 1436: TODO: Should check for readonly property and return PreAssignment only

PR #574, opened 2020-05-18
fix typo in readme

blubbll:
Link to issue(s) this covers

### Problem
typo (roylsn>roslyn)

### Solution
* [ ] At least one test covering the code changed


GrahamTheCoder:
I think we can waive the requirement for a covering test in this case, thanks 😊

PR #575, opened 2020-05-22
Fix race condition in GetExitCodeAsync().

dgrunwald:
After `WaitForExitAsync`, the git process has written all its output to the pipes and terminated.
However, it's possible that CodeConverter process has not yet read the data from the pipe, or that `OutputDataReceived` event is still executing after `GetExitCodeAsync()` returns.

This can cause an `ObjectDisposedException` in the `GetExitCodeAsync()` lambda expressions if the caller manages to dispose the `StringWriter` before all data is copied from the pipe to the `StringWriter`.

The `OutputDataReceived` event is called once with `e.Data == null` to signify the end-of-output. Waiting for that prevents the race condition.

### DIFF
--- a/CommandLine/CodeConv.Shared/Util/ProcessRunner.cs
+++ b/CommandLine/CodeConv.Shared/Util/ProcessRunner.cs
@@ -36,8 +36,20 @@ namespace ICSharpCode.CodeConverter.DotNetTool.Util
             psi.RedirectStandardError = true;
             psi.RedirectStandardOutput = true;
             using var process = new Process() { StartInfo = psi };
-            process.OutputDataReceived += (sender, e) => stdOut.WriteLine(e.Data);
-            process.ErrorDataReceived += (sender, e) => stdErr.WriteLine(e.Data);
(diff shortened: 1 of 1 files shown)


dgrunwald:
By the way, I've noticed you use `Microsoft.VisualStudio.Threading` for `WaitForExitAsync`. I have a simple implementation of this extension method (based on my old [SharpDevelop ProcessRunner](https://github.com/icsharpcode/SharpDevelop/blob/3f3ae2a5d4ffafd98cdf1c09308b9b8d93b0356c/src/Main/Base/Project/Util/ProcessRunner.cs#L549)) if you are interested in eliminating that dependency.

GrahamTheCoder:
Cheers for the PR, and for the information - I didn't know about them firing the event with null.

For the `Microsoft.VisualStudio.Threading` I'm happy to keep it. In fact I'd recommend that others check it out. It isn't actually tied to Visual Studio despite the naming. It contains a few helpful tools for managing async stuff such as `AsyncLazy`.

Thanks!

Issue #245, opened 2019-02-21
Add editor context menu to "Paste as VB"/"Paste as C#"


beppe9000:
I used to use the PasteAs extension, but it's for VS 2005, so it's outdated.
Does CodeConverter have that feature? I can't find it, so in the doubt I suggest it as a feature.

GrahamTheCoder:
There's an option you can enable so that the conversion result is copied to the clipboard:
https://github.com/icsharpcode/CodeConverter/issues/37#issuecomment-385810811
That might help you a bit.

However, I can see that if you want to copy something from a website like stack overflow and then paste it straight into your project of a different language, the feature you suggest could save some hassle.

I'm very happy to take a PR to add it. It should be reasonably easy to do - just add an extra menu item, register it in the command, and call the existing ConvertText method using the clipboard as source (copying from the way the other ones work). Are you interested in implementing it?

beppe9000:
I never worked with vs extensions, but can try :)

GrahamTheCoder:
Great 😄 I know it can be hard to figure out what's happening at the top level of a VS extension particularly, so here's a few pointers to relevant bits of code.

The most extension related part is duplicating the context menu entries in this file:
https://github.com/icsharpcode/CodeConverter/blob/master/Vsix/REConverterPackage.vsct#L31
https://github.com/icsharpcode/CodeConverter/blob/master/Vsix/REConverterPackage.vsct#L65

Then if you see where the existing command is used (note how the id defined matches the one defined at the bottom of the vsct file), it's a case of setting up your own handlers.
https://github.com/icsharpcode/CodeConverter/blob/e008f14d116bdb1619607c185cae8ab1e3e40627/Vsix/ConvertVBToCSCommand.cs#L23
https://github.com/icsharpcode/CodeConverter/blob/e008f14d116bdb1619607c185cae8ab1e3e40627/Vsix/ConvertCSToVBCommand.cs#L26

There's a handler for deciding if the command is visible, and one for doing the conversion (for each direction of language conversion):
https://github.com/icsharpcode/CodeConverter/blob/e008f14d116bdb1619607c185cae8ab1e3e40627/Vsix/ConvertVBToCSCommand.cs#L142
https://github.com/icsharpcode/CodeConverter/blob/e008f14d116bdb1619607c185cae8ab1e3e40627/Vsix/ConvertCSToVBCommand.cs#L145

You'll probably need to write a new public wrapper that calls through to this method in the CodeConversion class:
[...]

PR #622, opened 2020-08-31
Paste as VB/C# commands

BrianFreemanAtlanta:
Link to issue(s) this covers
This is a first attempt at #245. 

### Solution
I have pulled in code to add "Paste Special" to the editor pop up context menu. Currently the menu items appear in both the Paste Special menu and also the main edit menu under the "convert to VB" or "Convert to C#" menu items. For command visibility I used the existing EditorMenuVisibility (CS or VB) constraints. The converted code is added at the insertion point. It doesn't replace the current select (if code is selected).

**Current issues**:
* Debug session not consistently adding converted code
when I was debugging the first attempt to paste code in doesn't appear in the UI window. I'm not sure why at this point.
* Does not replace existing selection (do we need to worry about this?)
* Changelog not updated yet. Once questions/issues resolved I can update the changelog.

**Current Questions**:
* Should we keep menu items in both places ("edit" menu and "Paste special") or just paste special?
* Should we worry about the currently active selection or just insert at the cursor position?

I didn't add any test cases since it's for the add-in.


GrahamTheCoder:
Thanks, this looks like a great start. I'll try to have a more detailed look in the next few days and make some suggestions.

My instinct for the menu items is that both should exist so within a VB file you can either "Copy as C#" (use case: want to share some code on stack overflow but the question  requested CSharp) , or "Paste as VB" (use case: copied some stack overflow snippet in c# but project is VB).

GrahamTheCoder:
Looks great. I've connected it up to the progress window, and (I think) fixed a minor timing bug where it was pasting the text where their cursor was after conversion rather than at the time of clicking. Thanks for that!

Issue #625, opened 2020-09-10
Use Azure Functions Implementation of Code Converter


christophwille:
Available APIs:
* https://dotnet-codeconverter.azurewebsites.net/api/Version 
* https://dotnet-codeconverter.azurewebsites.net/api/Convert

This is hosted on the ic#code Azure account (not my personal one; it is a free one that comes with 1m calls). CORS is enabled.

What could we do with it? Tell users to use that backend when actually using the Web API (not the one that comes with the current Web frontend), or maybe switch the existing Web frontend to a pregeneration framework and host as static while calling this backend API.

![cconazure](https://user-images.githubusercontent.com/344208/92719965-3930cc80-f364-11ea-80d9-0e5c6498fdca.png)

GrahamTheCoder:
Might be worth trying hooking the ui to it and then publishing the ui as a static site via github pages.

GrahamTheCoder:
Theoretically the ui could have a dropdown to allow trying out different versions (mainly useful for reproducing regressions)

christophwille:
Yes, hosting statically on GH pages would be a good idea.

GrahamTheCoder:
I've started a draft PR #627 moving to a static site using the default VS template.
It needs more tidying, but runs locally now for me so I thought I'd mention it in case that unblocks any testing you wanted to do.
Note: I haven't hooked up to the azure function endpoint yet.

GrahamTheCoder:
I've now set the deployed azure function as the default endpoint for a release build (and it's currently running that way in production - though due to some odd caching, some users may need a hard refresh to get the newest version)

Issue #546, opened 2020-03-16
Improve web snippet conversion experience


GrahamTheCoder:
**Is your feature request related to a problem? Please describe.**
I just want to type and convert a quick snippet of code, but the converter throws an error in some cases when it's not neatly contained in a class. 

**Describe the solution you'd like**
Use an existing editor to help me write compiling input. 

**Describe alternatives you've considered**
The default template should have a function in it, and the cursor should start inside it.

**Requirements**
Must: still support basic conversion on mobile
Should: Show compile errors
Could: support intellisense/autofixups for using statements etc.

Saibamen:
Related to #445

mrmonday:
Might be cool to see https://microsoft.github.io/monaco-editor/ integrated with it.

GrahamTheCoder:
Yeah I think that and [SharpLab](https://github.com/ashmind/SharpLab) are the main two contenders to be compared. I like the autofixup options in SharpLab, but the same may be possible in monaco, I haven't investigated it. If anyone wants to do/find a comparison of the two and look for other options that'd be helpful.

Relatedly: I noticed recently that dotnetfiddle has a "Convert to VB/C#" option which uses this converter. I also notice SharpLab uses the ICSharpCode ILSpy library - https://github.com/ashmind/SharpLab/issues/444.
If our website became essentially a fork of SharpLab with the code converter, it might be worth checking if there's interest in integrating that feature to the core.

christophwille:
The current implementation is almost JAMstack (the pages use only the templating Razor feature...). We could go full JAMstack (https://www.staticgen.com/) and host the API eg in Azure Functions (published via a GH Action).

GrahamTheCoder:
**Further investigation**
Monaco only provides syntax highlighting out of the box. But to really tackle this issue we'll need a server to provide the roslyn suggestions : https://github.com/Microsoft/monaco-editor/issues/572
SharpLab is the same - they all run Roslyn like the converter.

So I think the easiest solution for this ticket would be using SharpLab.

**Static sites**
[...]

PR #627, opened 2020-09-13
Static app from default template

GrahamTheCoder:
We'd like to turn the UI into a static site that can be hosted on GitHub pages for simplicity, that calls azure functions to do the conversion.

In preparation for https://github.com/icsharpcode/CodeConverter/issues/546 and #625

WIP - I've picked the most commonly used of the default Visual Studio Single Page App templates and started porting the site to it.


christophwille:
I was thinking more of VS Code than VS. And using plain JS tooling.

GrahamTheCoder:
The (untested) intent was that the csproj is just a convenient way to integrate into the build pipeline and makes the debug button in vs work for a convenient dev experience alongside the other bits. Using vscode and running npm start in the client app folder should work just as well.

GrahamTheCoder:
I.e. I'm refactoring to something we could either deploy exactly as we do now, or repoint the api requests and deploy to github pages.

GrahamTheCoder:
I'm busy the next few days, but should be able to finish this next week

GrahamTheCoder:
I've added a readme explaining the options for running the site:
https://github.com/icsharpcode/CodeConverter/blob/master/Web/ClientApp/README.md

Let's discuss further development down this route on #625

PR #628, opened 2020-09-13
VB -> C#: Add brackets for null coalesce operator when needed

### LINKED ISSUE #626: VB -> C#: Null coalesce missing brackets
### Input code
```vbnet
Public Class VisualBasicClass
   Public Sub New()
       Dim y As String = If(("xxx")?.InnerText, String.Empty).ToUpper()
   End Sub
End Class
```

### Erroneous output
```C#
using System;

public partial class VisualBasicClass
{
    public VisualBasicClass()
    {
        string y = "xxx"?.InnerText ?? string.Empty.ToUpper();
    }
}
```

### Expected output
```C#
using System;

public partial class VisualBasicClass
{
    public VisualBasicClass()
    {
        string y = ("xxx"?.InnerText ?? string.Empty).ToUpper();
    }
}
```

### Details
* Because the parentheses are missing in the Erroneous output the value of .InnerText is not uppercased. But left in lowercase.
* Current online code converter at https://codeconverter.icsharpcode.net/ can be used to reproduce this issue.

SuperRembo:
Fixes #626 

### Problem
In some cases the null coalesce operator is missing parenthesis, resulting in incorrect output.

### Solution
* The same mechanism is used as already exists for the ternary operator
* Covered by test `NullCoalescingOperatorUsesParenthesisWhenNeededAsync`

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -611,11 +611,16 @@ namespace ICSharpCode.CodeConverter.CSharp
 
         public override async Task<CSharpSyntaxNode> VisitBinaryConditionalExpression(VBasic.Syntax.BinaryConditionalExpressionSyntax node)
         {
-            return SyntaxFactory.BinaryExpression(
+            var expr = SyntaxFactory.BinaryExpression(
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
Looks great, thanks!

PR #646, opened 2020-10-06
Service Worker from latest CRA

Saibamen:
`serviceWorker` from latest Create React App

Plus add new ignores to `.gitignore` (also from latest CRA)

### DIFF
--- a/Web/ClientApp/.gitignore
+++ b/Web/ClientApp/.gitignore
@@ -1,7 +1,9 @@
-# See https://help.github.com/ignore-files/ for more about ignoring files.
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
 # dependencies
 /node_modules
+/.pnp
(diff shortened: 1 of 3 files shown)


GrahamTheCoder:
Hi, thanks for the PR. I don't think this is the recommended way to update, I think it should usually be e.g.
`npm install --save --save-exact react-scripts@3.4.3`
See https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md

I don't think the service worker is used right now, so we should probably remove it. We don't really need the gitignore updates, might be worth consolidating with the main git ignore since only a couple of its entries are probably needed at all (other than node_modules)

GrahamTheCoder:
Thanks for the thought here. I'm going to close this and revisit the whole area of the service worker (and whether/how to best use it) at a later date since we'll need to figure out expectations and test cases

PR #656, opened 2020-10-14
Fix: Crash if relative path supplied to command line

leezer3:
### Problem
The command line converter only currently takes absolute paths, and crashes if a relative path is supplied.

### Solution
Check if the path is rooted & that the file does not exist.
If so, try combining with the current directory.

### DIFF
--- a/CommandLine/CodeConv.Shared/CodeConvProgram.cs
+++ b/CommandLine/CodeConv.Shared/CodeConvProgram.cs
@@ -120,13 +120,17 @@ Remarks:
 
         private async Task ConvertAsync(IProgress<ConversionProgress> progress, CancellationToken cancellationToken)
         {
+            string finalSolutionPath = SolutionPath;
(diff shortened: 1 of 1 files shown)


GrahamTheCoder:
Sounds like a very sensible pr, thanks! I'll try to get it tested and merged this weekend. I just want to check whether there is any validation built into the command line library being used that will either help or hinder here

GrahamTheCoder:
I've checked the FileExists attribute, which does something similar internally. Looks good to me, thanks a lot for the PR!

GrahamTheCoder:
I removed the extra exists check after merge, let me know if there was a specific case where you'd seen it was required

leezer3:
Not directly no, & it's likely good enough.

It was more of a belt + braces approach; Mono and stuff in the $PATH can sometimes give odd results.

(Quite how well this runs on Mono I haven't tested)

PR #660, opened 2020-10-21
cs -> vb: fix for several problems, and regressions

Alexilia:
### Problem
Visual Basic 11.0 does not support null conditional operations. (operator ?. in generated RaiseEvent accessor)
Visual Basic 11.0 does not support multiline string literals.

### Solution
Add LanguageVersion to ConversionOptions. It allows converter to generate code, which can be compiled for older versions.
Fixed RiseEvent for VB versions < 14
Fixed multiline string literals for VB versions < 14

More accurate convert lambda to avoid redundant generated __Assign method,
ObsoleteAttribute and Exception should be fully qualified in generated __Assign and __Throw methods (compilation error),
Fixed regression with redundant direct cast of Nothing to nullable type (compilation error)

all are with tests


CodeConverter/Shared/ConversionOptions.cs
```diff
@@ -6,6 +6,7 @@ namespace ICSharpCode.CodeConverter.Shared
     public class ConversionOptions
     {
         public string RootNamespaceOverride { get; set; }
+        public object LanguageVersionOverride { get; set; }
```
GrahamTheCoder, line 9: Is it possible to just set the `TargetCompilationOptionsOverride.ParseOptions.LanguageVersion` instead of adding a new property?

GrahamTheCoder:
Thanks for the pull request. I've made some minor comments.

Bearing in mind that VB 14 was released in January 2015, if possible I'd like to understand a bit more about the need for VB11/VB12 support since the burden of maintaining support for old versions of things is quite high.
From what I understand, one of the main uses of the C#->VB converter is to patch in bits of code to keep legacy systems running that aren't worth the investment to make bigger changes.
Are there big issues stopping people just incrementing the version number they compile against? What's your particular use case for this if you don't mind sharing?

CodeConverter/VB/CommonConversions.cs
```diff
@@ -322,6 +321,10 @@ public LambdaExpressionSyntax ConvertLambdaExpression(CSS.AnonymousFunctionExpre
             return CreateLambdaExpression(singleLineExpressionKind, multiLineExpressionKind, header, statements, endBlock);
 
         }
+        public bool IsVoidLambda(SyntaxNode node) {
```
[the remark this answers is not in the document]
GrahamTheCoder: I've pushed this change to the branch

CodeConverter/Shared/ConversionOptions.cs
Alexilia: Yes, it is. I have get confused with these options.

Alexilia:
We use the converter to convert our big demonstration application which contains small code examples. Conversion should be fully automatical to reduce maintaining of the demo application, because new examples are added and old are changing continuously.
Our products support targeting .Net Framework 4.5, because most users have it. So their application's could be compiled with Visual Studio 2012 and C#5 or VB11.

In common case conversion depends on input code. Old syntax will be translated into old ones, and new syntax will be translated into new ones. There was only problems with converter's generated syntax, which can't be controlled using code to convert.

GrahamTheCoder:
Thanks for sharing. Converting code samples is a great use case for the CS->VB conversion, and I see that why you'd want it to be highly compatible. I'll try to keep it in mind for future development.

Thanks for the pr!

Issue #658, opened 2020-10-19
VB -> C#: CInt wrongly converted to (int)


marcal:
Hello,

I have seen a problem with the conversion of the CInt.
Cint (99.99) gives 100
While (int) 99.99 gives 99

Marc

### Input code
```vbnet
 Dim do_Tmp As Double = 9999 / 100
        MessageBox.Show(CInt(do_Tmp))
```

### Erroneous output
```C#
double do_Tmp = 9999d / 100d;
        MessageBox.Show(((int)do_Tmp).ToString ()));
```

### Expected output
```C#


double do_Tmp = 9999d / 100d;
        MessageBox.Show((Conversions.ToInteger (do_Tmp).ToString ());
```

### Details
Last plugin version (.NET) and the online converter have the problem

GrahamTheCoder:
Hopefully it's just a case of making sure a conversion method is found somewhere around this point:
https://github.com/icsharpcode/CodeConverter/blob/a5473fbf194d70daa6b47a5267c22337a6eef2e7/CodeConverter/CSharp/ExpressionNodeVisitor.cs#L213-L217

GrahamTheCoder:
Fixed by #669

PR #669, opened 2020-10-25
VB -> C# explicit cast operations float/double/single to integer

BrianFreemanAtlanta:
Link to issue(s) this covers
This is WIP to address #658 . 

### Problem
In looking at this it I found that the system goes through `VisitPredefinedCastExpression`, `AddExplicitConversion`, `AnalyzeConversion` and finally into `TryAnalyzeCsConversion`.
currently the code below returns TypeConversionKind = NonDestructiveCast:
https://github.com/icsharpcode/CodeConverter/blob/9817cfc7733599ccf024d73d0273f609f6fb93b1/CodeConverter/CSharp/TypeConversionAnalyzer.cs#L279-L287
Which results in the (int) usage.

### Solution
* I added a check IsExplicitNumericConversion to test that we are using CInt. To return `TypeConversionKind.Conversion` for this case.
* The ConvertWholeSolution is failing. It shows that all the cast functions in Conversions class are also being updated. How can we exclude the Conversions class from this rule? I'm guess there may be something from the models that can be queries, but not sure what.

If you can help out with a pointer on what I might query or a different approach I can update the pull request.


GrahamTheCoder:
Thanks for the pr. Looks like it's pretty close. I don't think there's anything special about the conversions class (in terms of how it's compiled), so if it looks wrong it probably means there's a more specific method call that should be used in all cases. Decompliling the result of the VB compilation should give some hints hopefully. Let me know if you need more guidance or other ideas, thanks!

GrahamTheCoder:
I had a bit of a look and realized I don't think those conversion methods should actually be part of the output at all. I'll take a look at removing them some time.

For this PR, the code is still a good example showing this might not be the ideal conversion. I'll try to get more details soon on what the VB compiler does so we can decide the best next step.

BrianFreemanAtlanta:
When I did a web search I found this reference from VB.NET language in a Nutshell by Steven Roman PhD
https://www.oreilly.com/library/view/vb-net-language/0596000928/re40.html

> CInt differs from the Fix and Int functions, which truncate, rather than round, the fractional portion
This is confirmed I created a small VB application that had
```
  Sub Main()
        Dim doTmp As Double = 9999 / 100
        Console.Write(CInt(doTmp))
    End Sub
```
When I use ILDASM.exe on the exe I see the following code has been generated.

```
// Code size       25 (0x19)
    .maxstack  1
    .locals init ([0] float64 doTmp)
    IL_0000:  nop
    IL_0001:  ldc.r8     99.989999999999995
    IL_000a:  stloc.0
    IL_000b:  ldloc.0
    IL_000c:  call       float64 [mscorlib]System.Math::Round(float64)
    IL_0011:  conv.ovf.i4
    IL_0012:  call       void [mscorlib]System.Console::Write(int32)
    IL_0017:  nop
    IL_0018:  ret
```

Note there is a call to `System.Math::Round(Float64)` that VB has inserted before the conversion to integer.

Even though the suggestions was to change the translation to `Conversion.ToInteger` the current function doesn't provide the rounding functionality. In the few steps through I'm not catching when the "conversions" class is added to the compilation, but not sure if that's the right route or not.

GrahamTheCoder:
Good find

The "conversions" class gets added here:
https://github.com/icsharpcode/CodeConverter/blob/fb4b09a2f2e2515cc18a82678e0ab6196fe5cbfe/CodeConverter/CSharp/ProjectMergedDeclarationExtensions.cs#L75-L82
The code was intended to add in code that actually gets generated and compiled into the target assembly, but it seems it needs a filter to be applied to avoid that conversions class.

If you put a breakpoint around there and run the whole solution tests you'll be able to see the whole input vb. If the right path isn't obvious from that code, the next step would be to copy the code into a new project, compile, and decompile to C#. I was intending to do this just now, and post the results but the tests don't work for me since I installed .NET 5 and I haven't got around to fixing that yet!

BrianFreemanAtlanta:
I did a couple of thing in investigation. I downloaded DotPeek from JetBrains. A small exe with the lines
```
  Sub Main()
        Dim doTmp As Double = 9999 / 100
        Console.Write(CInt(doTmp))
    End Sub
```
Decompiles into
```
    public static void Main() => Console.Write(checked ((int) Math.Round(99.99)));
```
When I view Microsoft.VisualBasic.CompilerServices.Conversions for ToInteger this is what it has
```
    public static int ToInteger(object Value)
    {
      if (Value == null)
        return 0;
      if (Value is IConvertible convertible)
      {
        switch (convertible.GetTypeCode())
        {
          case TypeCode.Boolean:
            return Value is bool flag ? -(flag ? 1 : 0) : -(convertible.ToBoolean((IFormatProvider) null) ? 1 : 0);
          case TypeCode.SByte:
            return Value is sbyte num ? (int) num : (int) convertible.ToSByte((IFormatProvider) null);
          case TypeCode.Byte:
            return Value is byte num ? (int) num : (int) convertible.ToByte((IFormatProvider) null);
          case TypeCode.Int16:
            return Value is short num ? (int) num : (int) convertible.ToInt16((IFormatProvider) null);
          case TypeCode.UInt16:
            return Value is ushort num ? (int) num : (int) convertible.ToUInt16((IFormatProvider) null);
          case TypeCode.Int32:
            return Value is int num ? num : convertible.ToInt32((IFormatProvider) null);
          case TypeCode.UInt32:
            return Value is uint num ? checked ((int) num) : checked ((int) convertible.ToUInt32((IFormatProvider) null));
          case TypeCode.Int64:
            return Value is long num ? checked ((int) num) : checked ((int) convertible.ToInt64((IFormatProvider) null));
          case TypeCode.UInt64:
            return Value is ulong num ? checked ((int) num) : checked ((int) convertible.ToUInt64((IFormatProvider) null));
          case TypeCode.Single:
            return Value is float num ? checked ((int) Math.Round((double) num)) : checked ((int) Math.Round((double) convertible.ToSingle((IFormatProvider) null)));
          case TypeCode.Double:
            return Value is double a ? checked ((int) Math.Round(a)) : checked ((int) Math.Round(convertible.ToDouble((IFormatProvider) null)));
          case TypeCode.Decimal:
            return Value is Decimal ? convertible.ToInt32((IFormatProvider) null) : Convert.ToInt32(convertible.ToDecimal((IFormatProvider) null));
          case TypeCode.String:
            return Value is string str ? Conversions.ToInteger(str) : Conversions.ToInteger(convertible.ToString((IFormatProvider) null));
        }
      }
      throw new InvalidCastException(Utils.GetResourceString("InvalidCast_FromTo", Utils.VBFriendlyName(Value), "Integer"));
    }
```
Note for TypeCode.Single and TypeCode.Double it returns `(int)Math.Round(num)` and not just `(int)Num`
I also found that `Convert.ToInt32` is equivalent to `Cint` in VB with rounding applied. I think it may be a matter of tweaking `AddExplicitConvertTo` to invoke the proper Convert.ToXXX method instead of Conversions.ToXXX.

GrahamTheCoder:
In a sense the converter (like the compiler) is trying to Inline the relevant part of the method based on any type information known, then fall back (as it does now) on just calling the method of there's not enough compile time info.
Some interesting test cases might be in a constant context (e.g. Declaring a const field, or an optional parameter value) where in c# it'd only work as a cast.

GrahamTheCoder:
So perhaps just special case your fix to single and double input types will be enough for now?
Even if it causes a compile error for const I'd be happy to merge I think since it would currently be an invisible logic error

CodeConverter/Util/FromRoslyn/ITypeSymbolExtensions.cs
```diff
@@ -254,6 +254,36 @@ public static bool IsUnexpressibleTypeParameterConstraint(this ITypeSymbol typeS
...
+            }
+
+            return false;
+        }
+
+        public static bool IsIntegerNumericType(this ITypeSymbol? type)
+        {
+            if (type != null) {
+                switch (type.SpecialType) {
+                    case SpecialType.System_Int16:
```
GrahamTheCoder, line 275: Shouldn't this also include
```csharp
                     case SpecialType.System_Byte:
                    case SpecialType.System_SByte:
```

I'm not sure about char, the fact it isn't in the existing method below means it probably isn't required, but might be worth checking the behaviour.

GrahamTheCoder: If byte and sbyte are needed, then consider implementing some of these methods in terms of the others to avoid inconsistencies in future changes.
e.g. `IsNumericType(type) && !IsFloatingNumericType(type)`

CodeConverter/Util/FromRoslyn/ITypeSymbolExtensions.cs
```diff
@@ -254,6 +254,36 @@ public static bool IsUnexpressibleTypeParameterConstraint(this ITypeSymbol typeS
             return false;
         }
 
+        public static bool IsFloatingNumericType(this ITypeSymbol? type)
```
GrahamTheCoder, line 257: Perhaps a better name would be something like `IsFractionalNumericType` - since a decimal doesn't have a floating point this implementation surprised me

GrahamTheCoder:
This is looking in good shape now. I've added a couple of minor comments, but I'm happy to

BrianFreemanAtlanta:
I will update. It looks like Byte & SByte are needed, but is not applicable to Char.

GrahamTheCoder:
Sorry I've been too busy to be of more help, thanks for your great work on this, it's really appreciated!

BrianFreemanAtlanta:
I understand how it goes. We got to a solution 👍

PR #680, opened 2020-11-09
CS -> VB: improve generation of Call statement

Alexilia:
### Problem
there is no CallStatementSyntax for ObjectCreationExpression when several expressions are used. For example
new TaskFactory() as TaskFactory)
            .StartNew(() => { })
            .ContinueWith(t => {});

### DIFF
--- a/CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
+++ b/CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
@@ -67,22 +67,21 @@ namespace ICSharpCode.CodeConverter.VB
         {
             var exprNode = node.Accept(_nodesVisitor);
             var expression = exprNode as ExpressionSyntax;
-            if (expression != null) {
-                var allDirectExpressions = expression.FollowProperty(exp => exp.TypeSwitch(
+            return expression != null ? WrapRootExpression(expression) : (StatementSyntax)exprNode;
+        }
+        private StatementSyntax WrapRootExpression(ExpressionSyntax expressionSyntax) {
+            InvocationExpressionSyntax invocationExpression = expressionSyntax as InvocationExpressionSyntax;
+            if (invocationExpression != null) {
+                var lastExpression = expressionSyntax.FollowProperty(exp => exp.TypeSwitch(
                         (InvocationExpressionSyntax e) => e.Expression,
                         (ParenthesizedExpressionSyntax e) => e.Expression,
                         (CastExpressionSyntax e) => e.Expression,
                         (MemberAccessExpressionSyntax e) => e.Expression
-                    ));
-                InvocationExpressionSyntax invocationExpression = expression as InvocationExpressionSyntax;
-                var lastExpression = allDirectExpressions.LastOrDefault();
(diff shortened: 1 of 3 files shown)


CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
```diff
@@ -66,8 +66,19 @@ public override SyntaxList<StatementSyntax> VisitLocalDeclarationStatement(CSS.L
         private StatementSyntax ConvertSingleExpression(CSS.ExpressionSyntax node)
         {
             var exprNode = node.Accept(_nodesVisitor);
-            if (!(exprNode is StatementSyntax))
-                exprNode = SyntaxFactory.ExpressionStatement((ExpressionSyntax)exprNode);
+            var expression = exprNode as ExpressionSyntax;
+            if (expression != null) {
```
GrahamTheCoder, line 70: Nit: For cases like this, try to use the more modern syntax to be consistent with the rest of the codebase:
```csharp
if (exprNode is ExpressionSyntax expression)
```

CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
```diff
@@ -66,8 +66,19 @@ public override SyntaxList<StatementSyntax> VisitLocalDeclarationStatement(CSS.L
...
+                var allDirectExpressions = expression.FollowProperty(exp => exp.TypeSwitch(
+                        (InvocationExpressionSyntax e) => e.Expression,
+                        (ParenthesizedExpressionSyntax e) => e.Expression,
+                        (CastExpressionSyntax e) => e.Expression,
+                        (MemberAccessExpressionSyntax e) => e.Expression
+                    ));
+                if(allDirectExpressions.Last() is ObjectCreationExpressionSyntax)
```
GrahamTheCoder, line 77: I agree this is an improvement, but I don't think it covers all the cases. The [docs](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/call-statement#remarks) say
> You typically use the Call keyword when the called expression doesn’t start with an identifier. Use of the Call keyword for other uses isn't recommended.
So perhaps we could just check for something like:
`expression.DecendantsOrSelf<ExpressionSyntax>().FirstOrDefault() is IdentifierNameSyntax`

Sorry if I've misunderstood, maybe add a couple of other test cases?

CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
Alexilia: Unfortunately, modern syntax is acceptable only for boolean variables.

CodeConverter/VB/MethodBodyExecutableStatementVisitor.cs
Alexilia: I don't know another cases how c# code can be converted using Call statement. Example from [docs](https://docs.microsoft.com/en-gb/dotnet/visual-basic/language-reference/statements/call-statement)

`Call (Sub() Console.Write("Hello"))()`

can be converted from
`(() => Console.Write("Hello"))()`
but the C# code can't be compiled.
Anyway, I have prepared more common fix, but potentially it can add Call statement in excess cases.

GrahamTheCoder:
Sorry I forgot about this. Thanks!

Issue #684, opened 2020-11-28
VB -> C#: CS1009 Unrecognized escape sequence


AteCoder:
### Input code
```vbnet
$"Soak: {sender.SoakStopwatch.Elapsed:d\.h\:mm\:ss\.f}"
```

### Erroneous output
```C#
$"Soak: {sender.SoakStopwatch.Elapsed:d\.h\:mm\:ss\.f}";
```

### Expected output
```C#
$"Soak: {sender.SoakStopwatch.Elapsed:d\\.h\\:mm\\:ss\\.f}";
```

### Details
* Product in use: VS extension 
* Version in use: 8.2.1.0
* Did you see it working in a previous version, which? N/A
* Any other relevant information to the issue, or your interest in contributing a fix. N/A

GrahamTheCoder:
There's already some logic to escape normal strings, I expect the same logic can probably be applied to format strings too

PR #688, opened 2021-01-08
Add code for format clause escape character.

BrianFreemanAtlanta:
Link to issue(s) this covers
This closes issue #684 . 
### Problem
For format text on interpolated strings require an escape for the "\" character in C#, but not in VB. 

### Solution
* Added code to add an extra"\" converted to C# and to remove the unneeded "\" converting to VB.
* The test converting to VB is failing.  Would like pointer on getting it to pass. C# needs `using System;` but VB doesn't but the test says "comments for line 0 not converted." Without `using System;` the conversion is unsuccessful.
* [X] Covering unit tests: VB -> C# and C# -> VB

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -308,7 +308,8 @@ namespace ICSharpCode.CodeConverter.CSharp
 
         public override async Task<CSharpSyntaxNode> VisitInterpolationFormatClause(VBasic.Syntax.InterpolationFormatClauseSyntax node)
         {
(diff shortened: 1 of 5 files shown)


GrahamTheCoder:
Sorry, only just spotted this, thanks for the PR! Since I don't believe this introduces a new issue for the comment conversion I think it's fine to ignore it for the purpose of getting this change in. So I think if you put a comment on that line in the vb it won't get checked. e.g.
```vbnet
Imports System 'BUG: Code comment lost on this line
```

If that doesn't work, there's also a parameter to `TestConversionVisualBasicToCSharpAsync` which can disable testing comments entirely.

BrianFreemanAtlanta:
Thanks for the pointer. I updated the test case to ignore the comment issue. Should be good to go.

GrahamTheCoder:
Great, thanks! I'll try to release this in the next couple of weeks, but feel free to remind me if I forget - busy times!

Issue #682, opened 2020-11-22
C# -> VB: Generics type parameter missed BC30737 and BC32042


jonr3:
### Input code
```csharp
If relevant, please enter some example code here to reproduce the issue, or the steps taken to cause the issue.
```

### Erroneous output
[TaskCompletionSource2.zip](https://github.com/icsharpcode/CodeConverter/files/5579364/TaskCompletionSource2.zip)

```vbnet
            Dim tcs As TaskCompletionSource = Nothing

errors:

Severity	Code	Description	Project	File	Line	Suppression State
Error	BC32042	Too few type arguments to 'TaskCompletionSource(Of TResult)'.	TaskCompletionSource2	D:\clients\SADS\AsyncAwait\TaskCompletionSource2\TaskCompletionSource2\SdkProxy.vb	27	Active
```

### Expected output
```vbnet
I believe
Dim tcs As TaskCompletionSource(Of Boolean) = Nothing
```

### Details
* Product in use:  VS extension, VS2019 16.8.2
* Version in use: e.g. 8.2.1.0
* Did you see it working in a previous version, which? No, not used previous versions.
* Any other relevant information to the issue, or your interest in contributing a fix.

the Main sub is not correctly configured in the project so I also get this error:

Severity	Code	Description	Project	File	Line	Suppression State
Error	BC30737	No accessible 'Main' method with an appropriate signature was found in 'TaskCompletionSource2'.	TaskCompletionSource2	D:\clients\SADS\AsyncAwait\TaskCompletionSource2\TaskCompletionSource2\vbc	1	Active

whereas the c# code works fine with no changes.

GrahamTheCoder:
Thanks for the report. Both issues should be straightforward to fix, it's just a case of finding time. I'll prioritise the generics one since it can cause lots of work on a big project whereas there's only one main method

jonr3:
Thanks Graham,

The fix for the missing task type is:

Dim tcs As TaskCompletionSource(Of Boolean) = Nothing

Which compiles, as to whether its correct to allocate to nothing is another matter. The class TaskCompletionSource appears to require one and only one argument.

The duplicate “main” procedures are fixed by changing the private shared call to:

Public Shared Sub main(ByVal args As String())

It doesn’t seem to matter that the project points to the correct “main” as regardless of whether I choose “Sub Main” or TaskCompletionSource2.program the code works as it does in C# to the end.

[...]

PR #692, opened 2021-01-18
Updated ConvertToVariableDeclarator to handle generic declarations

BrianFreemanAtlanta:
Link to issue(s) this covers
This closes issue #682 .

### Problem
The system is generating a local variable for an out parameter. The type symbol was incorrectly parsed as a VB type which didn't properly recognize generics.

### Solution
* I replaced the call to `ToVBSyntax` with a call to the C# SyntaxFactory.ParseText instead of VB ParseText for the type symbol which then recognized the generic and the needed type arguments. I also added a call to the node visitor on the resulting CS type syntax.
* The ToVBSyntax had a NULL check (for type Symbol returned from `ExtractBestMatch<ITypeSymbol>`. Unsure if an explicit  NULL check and error throwing is needed
* Resharper formatted ExpressionsTest.cs resulting in several "minor" changes to the file. I can undo if needed.
* I did need to add the `hasLineCommentConversionIssue` to this test plus the existing `DeclarationExpressionAsync` test

I reviewed the callers for the `ToVBSyntax` which is called from the `ConvertToVariableDeclarator` overload for property declarations. It didn't seem to be an issue because it should never be called. A property cannot be declared as `var` so the call is unneeded. I believe the case will always fall into the `else` clause that simply calls the node visitor for the type.

### DIFF
--- a/CodeConverter/VB/CommonConversions.cs
+++ b/CodeConverter/VB/CommonConversions.cs
@@ -148,9 +148,7 @@ namespace ICSharpCode.CodeConverter.VB
             TypeSyntax typeSyntax;
             if (des.Type.IsVar) {
                 var typeSymbol = ModelExtensions.GetSymbolInfo(_semanticModel, des.Type).ExtractBestMatch<ITypeSymbol>();
-                var csTypeSyntax =
-                   CS.SyntaxFactory.ParseTypeName(typeSymbol.ToMinimalDisplayString(_semanticModel, des.Type.SpanStart));
-                typeSyntax = (TypeSyntax)csTypeSyntax.Accept(_nodesVisitor);
+                typeSyntax =(TypeSyntax) VbSyntaxGenerator.TypeExpression(typeSymbol);
             } else {
                 typeSyntax = (TypeSyntax)des.Type.Accept(_nodesVisitor);
             }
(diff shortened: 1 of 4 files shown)


GrahamTheCoder:
Sounds good. I'll try to take a closer look at the code and merge soon

CodeConverter/VB/CommonConversions.cs
```diff
@@ -142,12 +142,15 @@ public static StatementSyntax CreateLocalDeclarationStatement(SyntaxTokenList sy
...
+            
             var id = ((IdentifierNameSyntax)des.Accept(_nodesVisitor)).Identifier;
             var ids = SyntaxFactory.SingletonSeparatedList(SyntaxFactory.ModifiedIdentifier(id));
             TypeSyntax typeSyntax;
             if (des.Type.IsVar) {
                 var typeSymbol = ModelExtensions.GetSymbolInfo(_semanticModel, des.Type).ExtractBestMatch<ITypeSymbol>();
-                typeSyntax = typeSymbol?.ToVbSyntax(_semanticModel, des.Type);
+                var csTypeSyntax =
+                   CS.SyntaxFactory.ParseTypeName(typeSymbol.ToMinimalDisplayString(_semanticModel, des.Type.SpanStart));
+                typeSyntax = (TypeSyntax)csTypeSyntax.Accept(_nodesVisitor);
```
GrahamTheCoder, line 153: This class has a property called `VbSyntaxGenerator` which is a Roslyn class intended for converting a symbol to the appropriate syntax. Could you see if something like `VbSyntaxGenerator.TypeExpression(...)` achieves the same result

BrianFreemanAtlanta: I updated the code to use TypeExpression, replaced other call to ToVBSyntax and then removed ToVBSyntax.

GrahamTheCoder:
Great, thanks!

PR #693, opened 2021-01-21
VB -> C#: add async check for lamba expressions

### LINKED ISSUE #657: VB -> C#: Async Function doesn't translate to async lambda
### Input code
```vbnet
Return Await Me.ExecuteAuthenticatedAsync(Async Function()
    Return Await DoSomethingAsync()
End Function)
```

### Erroneous output
```C#
return await this.ExecuteAuthenticatedAsync(() => {
    return await DoSomethingAsync();
});
```

### Expected output
```C#
return await this.ExecuteAuthenticatedAsync(async () => {
    return await DoSomethingAsync();
});

```

### Details
* Version in use: 8.1.8.0

BrianFreemanAtlanta:
Link to issue(s) this covers
This closes #657 
### Problem
anonymous Async function without parameters was missing the async in converted code.

### Solution
* Added a check to see if anonymous function is Async like the other 2 lamba conversions
* I'm unsure if the else is needed (lines 56-61) there isn't a test that covers those lines, but since the other two routines check for anonymous I left it in. I think all lambas should be anonymous
* [X] Added test case to improve code coverage of lambda conversions

### DIFF
--- a/CodeConverter/CSharp/LambdaConverter.cs
+++ b/CodeConverter/CSharp/LambdaConverter.cs
@@ -47,11 +47,20 @@ namespace ICSharpCode.CodeConverter.CSharp
             var body = (CSharpSyntaxNode)block ?? expressionBody;
 
             LambdaExpressionSyntax lambda;
-            if (param.Parameters.Count == 1 && param.Parameters.Single().Type == null) {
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
Thanks!
I've slightly deduped the logic in 929585303a0d226076a4fb3538a1d89d9d564f69

PR #706, opened 2021-02-21
Build GH Action Build matching Azure DevOps Pipeline

christophwille:
I have not matched https://github.com/icsharpcode/CodeConverter/blob/master/azure-pipelines.yml#L57 (necessary???)

First run artifacts:

https://github.com/icsharpcode/CodeConverter/actions/runs/586676519#artifacts

Note that GH re-zips automatically for download (no way around this behavior) and that artifacts can be downloaded only when you are logged in to GH.

Build passing badge can be built for master/succeeded once merged. TODO after merge: delete azure-pipelines.yml & pipeline on Azure DevOps.

Please squash-merge.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: windows-latest
     
     env:
-      CC_BUILD_VERSION: '8.2.2.${{ env.Project_Path }}'
+      BuildVersion: '8.2.2.${{ env.Project_Path }}'
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


GrahamTheCoder:
Looks good. Just the code conv tool nuget package missing that I can see. The runsettings from the line mentioned just set up the code coverage to be a vaguely useful number so we can see if things are going down hill (not that I've ever really checked it)

christophwille:
https://github.com/icsharpcode/CodeConverter/blob/master/azure-pipelines.yml#L62 ah, that mislead me into thinking it wasn't packed and wouldn't need to be collected. Wrong.

screenshot from Azure DevOps

christophwille:
Now we got all three: https://github.com/icsharpcode/CodeConverter/actions/runs/588190266#artifacts

GrahamTheCoder:
@christophwille Before I delve into it, any guesses on why this is having issues now: https://github.com/icsharpcode/CodeConverter/runs/2264078251

christophwille:
That one issue we already ran into with ILSpy; however, the restore problem... I don't know (yet; and I am not sure if it is really 100% a gh actions issue).

GrahamTheCoder:
Thanks that's definitely helped. I think the remaining issue is now because vs isn't on the machine anymore, or the wrong version

christophwille:
VS is on the machine, but it is now 16.9 (it was 16.8 previously - given the messages in the actions log something did change)

PR #721, opened 2021-04-19
Projectnames with a suffix equal to other projects unjustly converted.

StealthyDeveloper:
Link to issue(s) this covers
There is none.

### Problem
Suppose we have project A and project B.A. When converting only project A the CodeConverter replaces the extension B.A.<csproj/vbproj> also, because it matches the regex search on A.<csproj/vbproj>.

### Solution
I've included the proj dir name in the regex search, since the combination of dir  name and proj name must be unique within a solution. I've also used this combo for the regex match for the proj guid replacement, since this match is also based upon project name only. 

For the Unit Tests I've added a new test project Prefix.Vblibrary to the source files, recharacterized and ran all the MultiFileSolutionAndProjectTests. The test fails before the code change and passes after.


CodeConverter/Shared/SolutionConverter.cs
```diff
@@ -112,9 +120,9 @@ private ConversionResult ConvertSolutionFile()
             };
         }
 
-        private static (string Find, string Replace, bool FirstOnly) GetProjectGuidReplacement(string projFilename, string contents)
+        private static (string Find, string Replace, bool FirstOnly) GetProjectGuidReplacement(string projPath, string contents)
         {
-            var projGuidRegex = new Regex(projFilename + @""", ""({[0-9A-Fa-f\-]{32,36}})("")");
+            var projGuidRegex = new Regex(projPath + @""", ""({[0-9A-Fa-f\-]{32,36}})("")");
```
GrahamTheCoder, line 125: Will this still work for projects directly in the solution directory? Perhaps we can extract this static method to somewhere like ProjectFileTextEditor, and run unit tests on it more like https://github.com/icsharpcode/CodeConverter/blob/master/Tests/LanguageAgnostic/ProjectFileTextEditorTests.cs

It might be that just prefixing the regex to look for any of `/ \ "` is sufficient for this example. Though I can certainly see that someone could name a project the same and put it in a different directory too, which the current solution handles better. Since there are a few possible cases I think unit tests will help a lot

GrahamTheCoder:
Thanks for the PR, looks good. Let me know if you're able to look at unit testing a few more cases around it

Tests/TestRunners/MultiFileTestFixture.cs
```diff
@@ -37,7 +37,7 @@ public sealed class MultiFileTestFixture : ICollectionFixture<MultiFileTestFixtu
         /// <summary>
         /// Turn it and run the test, then you can manually check the output loads/builds in VS.
         /// </summary>
-        private readonly bool _writeAllFilesForManualTesting = false;
+        private readonly bool _writeAllFilesForManualTesting = true;
```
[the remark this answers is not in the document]
StealthyDeveloper: Ah yes. I turned it on to add the new Prefix.Vblibrary project to the expected solution files in the characterization tests, but it added all the other project files from the sourcefiles as well. I'll manually delete the unnecessary ones again, since there is some perfomance penalty indeed and for the tests to be more concise/clear.

CodeConverter/Shared/SolutionConverter.cs
StealthyDeveloper: I think for projects inside the sln dir this will fail. I'll change it to use a relative path instead. 

Yes, the case you're describing is why I chose to include the dirname as well instead of just the dir separator.

GrahamTheCoder:
@StealthyDeveloper I've pushed some unit tests to the branch showing some cases that don't currently work. In fairness, I don't think those cases work on master currently either. I'll leave it for a bit in case you want to have a look, otherwise I'll try to continue it myself.

Looks like you have some interesting other branches in progress too 😃

StealthyDeveloper:
> @StealthyDeveloper I've pushed some unit tests to the branch showing some cases that don't currently work - feel free to improve them however you like. In fairness, I don't think those cases work on master currently either. I'll leave it for a bit in case you want to have a look, otherwise I'll try to continue it myself.
> 
> Looks like you have some interesting other branches in progress too 😃

Thanks! I didn't get around to revisiting this yet. I'll have a look this weekend and also set out pull requests for the other issues 👍

StealthyDeveloper:
I've expanded upon your UT's. I fixed them by using the proj path relative to the sln folder and by matching on the entire proj refererence line for the sln file and adding an extra replacement for the guid. For the proj file replacements the regex matches the character before the path as well.

GrahamTheCoder:
Thanks for this. Looks good!

I noticed just after merging that it makes the Multi File tests [fail on the build server](https://github.com/icsharpcode/CodeConverter/runs/2811226525) when they were working just before due to the other PR I'd just merged. Any guesses why this might be, I assume they pass locally for you? (as they do for me)

StealthyDeveloper:
I looked at your attempts and also locally. The nullreference which you got should be fixed now. I forgot to reset the TextReplacementConverter.FileSystem. For the initial problem about the guid I've thrown a more detailed exception to investigate the build on the server. See #735. 

The tests do pass locally for me as well, so I'm very curious.

PR #735, opened 2021-06-13
Fix failing Multifile tests on build server

StealthyDeveloper:
### Problem
Multifilecharacterization tests failing on buildserver.

Update: Dispose doesnt get called after a xunit test immediately.

### Solution
Reset TextReplacement.FileSystem in tests also.
throw custom exception to investigate projguid method parameters on server.

 Update: The FileSystem used doesn't share state anymore.


GrahamTheCoder:
Makes things simpler and makes the tests pass, good stuff, thanks!

Issue #720, opened 2021-04-19
VB -> C#: DataTable access by index wrong conversion


StealthyDeveloper:
### Input code
```vbnet

Private _myTable As DataTable

CInt(_myTable(0)("MyColumnName"))
```

### Erroneous output
```C#

private DataTable _myTable;

Conversions.ToInteger(_myTable.ElementAtOrDefault(0)["MyColumnName"])
```
### Expected output
```C#

private DataTable _myTable; 

Conversions.ToInteger(_myTable.Rows[0]["MyColumnName"])
```

### Details
* Product in use: VS extension
* Version in use: 8.2.3.0
* Not seen a fix in older versions.
* If possible, could there be a fix in a few days? Otherwise I'd be willing to contribute, seeing as the conversion can't really wait any longer.

GrahamTheCoder:
Sorry for the delay getting back to you. Hope the conversion went well if it went ahead!

When I try to compile your input code it issues an error for that one: [sharplab example](https://sharplab.io/#v2:DYLgbgRgNAJiDUAfAkgWwA4HsBOAXAzgAQDKAnvrgKaoB0AIgIa4MCwAUOwAoCuEwAlgGNCAYWAN8RZJO6UA7ACYADO0JrCnbPzBNKJABYNslGIQD6qUgBUGfPQEEijZjbur1m7bsIAPQgF5RZAA7XAAKC2tbYEowpQBKMIAiAFlSEUxgblRggDkGVEok+Pj2AFFg0zEJfCA)

Presumably DataTable is one of your own custom types Rows as the "default property", rather than  `System.Data.DataTable` that I'm trying to compile against?

StealthyDeveloper:
No problem I've been busy as well and implemented a workaround for this in the source code. The exact DataTable type being used originates from System.Data, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089, .NET Framework v4.8. This code compiles for me in LINQPad5: [http://share.linqpad.net/79upfh.linq](url) which uses v.4.7.2. 

What's even more interesting is that hovering over the "indexer" in LPad also shows ElementAtOrDefault. Looking into System.Data.DataTable the default property is TableName not Rows oddly enough.

GrahamTheCoder:
[...]

PR #744, opened 2021-08-02
Issues/indexing datatables

StealthyDeveloper:
### Problem
https://github.com/icsharpcode/CodeConverter/issues/720

### Solution
Added AsEnumerable() invocation expression for datatable invocations in ExpressionNodeVisitor.ConvertInvocationAsync
Added UT: AccessExpressionTests.DataTableIndexingAsync()


GrahamTheCoder:
Thanks for this! I'll try to take a look in the next few days.

CodeConverter/Shared/DefaultReferences.cs
```diff
@@ -29,8 +29,7 @@ public static class DefaultReferences
             typeof(System.Web.HttpUtility),
             typeof(System.Xml.XmlElement),
             typeof(System.Xml.Linq.XElement),
-            typeof(Microsoft.VisualBasic.Constants),
-            typeof(System.Data.SqlClient.SqlCommand),
+            typeof(Microsoft.VisualBasic.Constants)
```
GrahamTheCoder, line 32: ANy reason to remove SqlCommand?

CodeConverter/Shared/VisualBasicCompiler.cs
```diff
@@ -52,6 +52,8 @@ private static VisualBasicCompilation CreateVisualBasicCompilation()
 
         public static VisualBasicCompilationOptions CreateCompilationOptions(string rootNamespace = null)
         {
+            // Caution: The simplifier won't always remove imports unused by the code (e.g. System.Data)
+
```
GrahamTheCoder, line 56: One case that causes them not to be removed is when they can't be resolved - could this be to do with removing the reference to SqlCommand?

GrahamTheCoder (approved):
Looks good, just the one thing to check commented above.

Thanks for the PRs, they look great. I'm also all for using newer, terser syntax. But could I ask that in future PRs you try to separate formatting/tidying changes from behaviour changes?

CodeConverter/Shared/DefaultReferences.cs
StealthyDeveloper: Good catch. I initially thought it was unnecessary because of system.data.datarow but I see sqlcommand has been moved to its own assembly in later frameworks.

CodeConverter/Shared/VisualBasicCompiler.cs
StealthyDeveloper: I will check again tomorrow, but I don't think so because on .net472 (framework version of tests assembly) sqlcommand and datarow are within the same assembly. I suspect it's a name/overload conflict being resolved by the compiler between system.data and another assembly (possibly system.linq?) since if I copy the output of a failing test which gets an extra system.data using, the compiler itself doesn't trigger the usual unused using warnings.

CodeConverter/Shared/DefaultReferences.cs
StealthyDeveloper: Good catch. I initially thought it wasn't necessary with datarow just above, but I see that sqlcommand has been moved to its own assembly in other framework versions. 

> Looks good, just the one thing to check commented above.
> 
> Thanks for the PRs, they look great. I'm also all for using newer, terser syntax. But could I ask that in future PRs you try to separate formatting/tidying changes from behaviour changes?

Yes good call and sorry about that, that's easier to review.

CodeConverter/Shared/VisualBasicCompiler.cs
StealthyDeveloper: It was indeed caused by overloading resolution. I've updated the comment for clarification.

PR #746, opened 2021-08-03
Issues/importing types modules not converted

StealthyDeveloper:
### Problem
Class/Module imports are not converted correctly. They are converted literally (imports => using) to using statements which results in compiler errors.

### Solution
Use the c# static qualifier for these cases.
Added multicharacterization tests.


CodeConverter/CSharp/CommonConversions.cs
```diff
@@ -701,7 +766,19 @@ public static ExpressionSyntax NothingComparison(ExpressionSyntax otherArgument,
...
+
+        public async Task<string> GetClassificationLastTokenAsync(VBSyntax.SimpleImportsClauseSyntax clause)
+        {
+            // Global imports aren't associated with a document and no need to classify aliases
+            if (clause.SyntaxTree.FilePath == string.Empty || clause.Alias != null) { return null;}
+
+            var span = clause.GetLastToken().Span;
+            var spans = await Classifier.GetClassifiedSpansAsync(Document, span);
+
+            return spans.Last().ClassificationType;
```
GrahamTheCoder, line 781: This is cool, I didn't know about it and probably could have used it several other places, nice!

GrahamTheCoder (approved):
This looks good. I'll just revert the bunch of easily separated line-wrap formatting changes before I merge. Nice one!

PR #759, opened 2021-08-29
Allow app to use different api sources

GrahamTheCoder:
Allows `data-apisource` on the head element, or suffixes for the site which point it to different APIs:
`?apiSource=HostedFunc`
and
`?apiSource=LocalFunc` (only relevant for local debugging)

This means:
* Locally it's easy to test against the local/remote azure function without restarting the app
* The static app could trivially be published to github pages and pointed towards the hosted azure function


Func/Convert.cs
```diff
@@ -18,7 +18,7 @@ public static class Convert
         //
         [FunctionName("Convert")]
 #pragma warning disable VSTHRD200 // Use "Async" suffix for async methods - Name must be "Run" for this to work AFAIK
-        public static async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = null)] HttpRequest req, ILogger log)
+        public static async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] HttpRequest req, ILogger log)
```
GrahamTheCoder, line 21: @christophwille I was a bit confused when I came across this. It looks like it was currently a get request that has a body (which is weird). But even weirder is that the deployed version actually 404s for "GET", but works for "POST". So I don't think I'm changing anything, but just wanted to flag this in case I've misunderstodd what's going on here

Example url I tried with get/post:
https://dotnet-codeconverter.azurewebsites.net/api/Convert

Web/Controllers/ConvertController.cs
```diff
@@ -5,7 +5,7 @@
 namespace ICSharpCode.CodeConverter.Web
 {
     [Route("api/[controller]")]
-    public class ConverterController : Controller
+    public class ConvertController : Controller
```
GrahamTheCoder, line 8: I've changed this (and therefore the path) to be consistent with the Azure function.
If anyone is using this we could make the old path a redirect perhaps. Any thoughts?

Web/Controllers/ApiController.cs
```diff
@@ -0,0 +1,22 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace ICSharpCode.CodeConverter.Web.Controllers
+{
+    [Route("api")]
+    public class ApiController : Controller
+    {
+        [HttpPost]
+        [Route("Convert")]
+        [Route("Converter")] //Old name, kept for compatibility
```
GrahamTheCoder, line 13: I wasn't sure if anyone uses this existing API endpoint within the app so have left it here for now.

Func/Convert.cs
christophwille: POST was very intentional because of size limitations on GET via URL (and caching of proxies in between)

GrahamTheCoder: Ah phew, nothing weird going on here. It looks like it was accidentally changed to `GET` by my commit just before - I think I must have copy pasted when suppressing the naming issue and not noticed the difference 😬 All fixed now 😃

Issue #760, opened 2021-09-03
[Microsoft] Disabling Your Incompatible VS 2022 Extension


leslierichardson95:
Hi,
 
This is Leslie from the Visual Studio Extensibility team. I’m reaching out to you because we’ve received the following, high-priority feedback ticket for VS 2022: [Visual Studio Feedback](https://developercommunity.visualstudio.com/t/Could-not-load-file-or-assembly-EnvDTE/1463536)
 
We determined that this issue is the result of extensions not being updated to Visual Studio 2022 properly due to marking VS 2022 as compatible in the VSIX manifest but not referencing the new SDK PIA (Microsoft.VisualStudio.Interop.dll) and dependencies.  We’ve identified that the CodeConverter extension contributes to this issue.
 
Starting on **Tuesday, September 7th**, because of this issue’s severity, we are no longer allowing users to successfully install this extension until you **update and re-publish your VSIX with the new VS 2022-compatible references, including EnvDTE**.  
 
When re-publishing your extension, the existing VSIX validator will be expanded to check if that extension depends on the old assembly or not.  If the extension still uses these old dependencies and PIAs, the extension will not be published.

Q: How can I validate whether my extension has dependencies on the old PIAs assembly?
A: You can try to validate the extension’s reference by following the steps below

1. Open the Developer command prompt.
2. Run the command [ildasm](https://docs.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022) as below.
    - C:\>ildasm /text "{Extension Assembly}.dll"  /output:result.txt
3.	Check for references in the result.txt file.

For more info on how to migrate your extension to VS 2022 successfully, please view the following resources:

- [Update Your Visual Studio Extension](https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022)
- [Breaking API List](https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/breaking-api-list?view=vs-2022)
- [Migrated PIA List](https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/migrated-assemblies?view=vs-2022)
 
Please email me if you have any questions and sorry for the inconvenience.
 
Leslie Richardson
Program Manager – Visual Studio
leslie.richardson@microsoft.com | [@lyrichardson01](https://twitter.com/lyrichardson01) | [LinkedIn](https://www.linkedin.com/in/leslie-richardson-95a498b4/)

[...]

PR #761, opened 2021-09-05
Experiment for issue #760

christophwille:
Entirely removing PIA ref as per https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/migrated-assemblies?view=vs-2022 see issue #760


GrahamTheCoder:
Looks good to me, I tested Opening VS with a solution file, a conversion of some kind, opening the dialog page and closing VS against:
* 15.9.38
* 16.9.3
* 17.0.0 Preview 3.1

PR #762, opened 2021-09-09
Updated all dependencies and fixed invalid cast exception

Havunen:
Link to issue(s) this covers

### Problem
Updated all dependencies to latest version and target framework to 4.7.2
Fixes an invalid cast exception when converting optional parameters from VB to C#

Not all tests are passing on my machine, not sure if they did before. This PR could be used as basis if you want to update the dependencies

### DIFF
--- a/CodeConverter/CSharp/DeclarationNodeVisitor.cs
+++ b/CodeConverter/CSharp/DeclarationNodeVisitor.cs
@@ -1407,7 +1407,14 @@ namespace ICSharpCode.CodeConverter.CSharp
             var attrListsToRemove = parameter.AttributeLists.SingleOrDefault(aList => aList.Attributes
                .All(a =>
                 {
-                    var attrIdentifier = ((IdentifierNameSyntax)a.Name).Identifier.Text;
+                    string attrIdentifier = string.Empty;
+
+                    if (a.Name is IdentifierNameSyntax identifierNameSyntax) {
+                        attrIdentifier = identifierNameSyntax.Identifier.Text;
+                    } else if (a.Name is AliasQualifiedNameSyntax aliasQualifiedNameSyntax) {
+                        attrIdentifier = aliasQualifiedNameSyntax.Alias.Identifier.Text;
+                    }
+
                     return optionalAttributes.Contains(attrIdentifier);
                 }));
 
--- a/Vsix/Vsix.csproj
+++ b/Vsix/Vsix.csproj
@@ -3,6 +3,9 @@
   <PropertyGroup>
     <MinimumVisualStudioVersion>16.0</MinimumVisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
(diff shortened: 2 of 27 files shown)


CodeConverter/CSharp/DeclarationNodeVisitor.cs
```diff
@@ -1407,7 +1407,14 @@ private static ParameterSyntax ConvertOptionalParameter(ParameterSyntax paramete
             var attrListsToRemove = parameter.AttributeLists.SingleOrDefault(aList => aList.Attributes
                .All(a =>
                 {
-                    var attrIdentifier = ((IdentifierNameSyntax)a.Name).Identifier.Text;
+                    string attrIdentifier = string.Empty;
+
+                    if (a.Name is IdentifierNameSyntax identifierNameSyntax) {
```
Havunen, line 1412: this is the bug fix to invalid cast exception

Vsix/Vsix.csproj
```diff
@@ -3,6 +3,9 @@
   <PropertyGroup>
     <MinimumVisualStudioVersion>16.0</MinimumVisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
```
GrahamTheCoder, line 7: Purely out of interest, what method/tool did  you use to update things that caused this?

Havunen: visual studio 2022, created a new Vsix project because it did not compile for some reason

CodeConverter/CSharp/DeclarationNodeVisitor.cs
GrahamTheCoder: Great, please could you add a test case that fails before this change, and passes after.
Or if you're not sure how, let me know an example code snippet (probably includes an attribute) that caused this and I can help with that.

GrahamTheCoder:
Thanks for the PR. Yes, all the tests pass on master - they run in github actions too which backs the "passing" badge at the top of the readme to make sure we never accidentally break things.

I'm afraid the dot net dependencies on this project are reasonably finely tuned for maximum compatibility with VS2017-VS2022, so we'll need to undo those.

Npm updates are generally a good thing, so long as the website still behaves as expected.

I'm general it's best to split different unrelated  changes into different PRs so each can be tested/reviewed/merged independently.

CodeConverter/CSharp/DeclarationNodeVisitor.cs
Havunen: ok, I can try

Vsix/Vsix.csproj
GrahamTheCoder: Interesting, I use Vs2022 with no problem. Could you let me know what the compile error was?

Havunen:
I will open new PR with only test + bugfix

PR #763, opened 2021-09-12
Fixes an issue where Visual Basic optional parameter fails conversion to C#

Havunen:
### Problem
Fixes an invalid cast exception when converting optional parameters from VB to C#

### Solution

### DIFF
--- a/CodeConverter/CSharp/DeclarationNodeVisitor.cs
+++ b/CodeConverter/CSharp/DeclarationNodeVisitor.cs
@@ -1407,7 +1407,14 @@ namespace ICSharpCode.CodeConverter.CSharp
             var attrListsToRemove = parameter.AttributeLists.SingleOrDefault(aList => aList.Attributes
                .All(a =>
                 {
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
Awesome, thanks!

CodeConverter/CSharp/DeclarationNodeVisitor.cs
```diff
@@ -1407,7 +1407,14 @@ private static ParameterSyntax ConvertOptionalParameter(ParameterSyntax paramete
             var attrListsToRemove = parameter.AttributeLists.SingleOrDefault(aList => aList.Attributes
                .All(a =>
```
GrahamTheCoder, line 1408: Not caused by your PR, but just in passing I've noticed that this `All` looks suspicious, it makes me think that the individual attribute won't be removed if it's part of another attribute list e.g.:
```vbnet
Public Class MyTestAttribute
    Inherits Attribute
End Class
Public Class MyController
    Public Function GetNothing(
        <OptionalAttribute(), MyTest()> Optional indexer As Integer? = 0
    ) As String
        Return Nothing
    End Function
End Class
```

PR #768, opened 2021-09-26
Upgrade to dotnet 5

GrahamTheCoder:
The idea is to get on to a single set of build tools to:
* Simplify the project requirements
* Smooth out the VS experience
* Prepare for dot net 6

I'm using this as inspiration:
* https://stackoverflow.com/a/58947835/1128762
* https://github.com/VsVim/VsVim/blob/master/Src/VsVim2022/VsVim2022.csproj

No luck so far - it doesn't deploy the vsix on build. I'm going to wait a bit and hope Microsoft officially supports this properly


GrahamTheCoder:
For moving to dot net 6 we'll need to check the experience for people who don't have dot net 6 runtime installed to gauge when to move

Issue #787, opened 2021-11-09
Web + Func update to .NET 6


christophwille:
A simple tfm change might not be the best way to do this - both for the Azure Function as well as the Web project. For Azure Functions:

https://techcommunity.microsoft.com/t5/apps-on-azure/azure-functions-4-0-and-net-6-support-are-now-generally/ba-p/2933245

For the Web, SPA-integration has changed considerably, so a File New/Project and copy/paste from the old to the new might be best.

https://docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-6.0?view=aspnetcore-6.0#miscellaneous

christophwille:
Video on the new frontend templates from dotnetconf: https://www.youtube.com/watch?v=Zu6lui-rcOY

PR #788, opened 2021-11-18
Minimal update to .NET 6.0

christophwille:
...for Web, Function and added TFM for dotnet tool

This is the minimal move to .NET 6 - see #787 (Web should be re-done & copy/pasted into new from template)

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Setup .NET for main build
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.x
+        dotnet-version: 6.0.x
     - name: Add msbuild to PATH
       uses: microsoft/setup-msbuild@v1.0.2
       with:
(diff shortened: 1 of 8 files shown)


christophwille:
@GrahamTheCoder why do we do a "publish all"? The question is because we do a dotnet pack separately too, so shouldn't we be packing the codeconv tool & publishing func+web separately? (vsix doesn't need publish)

GrahamTheCoder:
> why do we do a "publish all"

Because that's conceptually what we're trying to do - get everything in a publishable state. e.g. nupkg / vsix / hostable folder

> Try to split build and publish

I don't mind the steps being split if it helps understand the build output, but in doing so is there any reason you've also moved from using the new `dotnet` command for the majority of the building to the old `msbuild` for everything instead? At some point I think this will cause us issues, but they'll likely be fairly noticeable, so we can stick with it until that point if you like. Just seems a bit backwards for a PR to do with using the latest versions.

christophwille:
Yes, using msbuild for everything is intentional (and no, it is not going away anytime soon). One reason is that https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation#deterministic needs it.

PR #789, opened 2021-11-18
Npm updates

GrahamTheCoder:
Tries to get to the latest version of all npm packages

Needed a prerelease of React-scripts since it pulls in a webpack version not compatible with the latest node (17+):
* https://github.com/webpack/webpack/issues/14532

### npm audit stuff
Also made sure a few things that got in the dependencies category go into devDependencies.
Sadly npm install will still complain about irrelevant vulnerabilities, but `npm audit --production` will then only tell us about the 0 real issues  - perhaps we could even make it part of the github actions build.
See https://github.com/facebook/create-react-app/issues/11174 for more info

### todo

* [ ] Remove newline from styling of header links


GrahamTheCoder:
Because of the vulnerabilities in existing packages, I'm going to merge this and tidy up the visual artifacts separately

PR #794, opened 2021-11-28
Switch to new spaproxy template from dotnet 6

### LINKED ISSUE #787: Web + Func update to .NET 6

GrahamTheCoder:
Fixes #787

When running in visual studio, the Web project is now really just a (semi-)convenient debug tool to:
* Start the webpack dev server
* Host a version of the API

When published, and run in IIS it:
* Hosts the built static files
* Host a version of the API

### Before merge 

* [x] Ensure both nuget packages get 3 ticks in nuget package explorer
image


christophwille:
Yes, we can host on GH pages, might actually be a good idea (alternatively: Azure Static Web Apps, we do that with a custom copy of release stats here https://github.com/christophwille/github-release-stats)

GrahamTheCoder:
I've just tested it in IIS and it does work by magic (i.e. it doesn't try to run the webpack dev server)
So we can still do either mode as before

GrahamTheCoder:
Azure static app sounds good (in some ways the github pages approach of having a branch in the repo with the built code is a bit ugly)

christophwille:
If we publish to whereever - should that on "release"? Something like https://github.com/icsharpcode/ilspy-vscode/blob/master/.github/workflows/ci.yml#L46 or more elaborate https://github.com/icsharpcode/SharpZipLib/blob/master/.github/workflows/release.yml

GrahamTheCoder:
Ah yeah that'd be really cool. I've created an issue for it here: https://github.com/icsharpcode/CodeConverter/issues/795

PR #796, opened 2021-12-01
Make it look more like a minimal API project

christophwille:
@GrahamTheCoder please have a look if this still correctly works in all CORS scenarios.


GrahamTheCoder:
CORS is just for running locally - because since the last PR, the API is on a different port to the UI

Issue #623, opened 2020-09-06
VB -> C#: Static variables in VB are not converted to C#


rlktradewright:
### Input code
```vbnet
    Private Sub IStrategyHostController_NotifyPosition(pPosition As Integer) Implements IStrategyHostController.NotifyPosition
        Static sPrevPosition As Integer

        mView.NotifyPosition(pPosition)

        If (pPosition <> 0 And sPrevPosition = 0) Or (pPosition > 0 And sPrevPosition < 0) Or (pPosition < 0 And sPrevPosition > 0) Then
            If mModel.IsTickReplay Then
                mView.EnableTradeDrawing()
                mView.DisableTradeDrawing()
            End If
            mTradeBarNumber = mTradeBarNumber + 1
            If mModel.ShowChart Then
                Logger.Log(LogLevels.LogLevelNormal, $"New trade bar: {mTradeBarNumber} at {mModel.Ticker.Timestamp}")
                mView.NotifyNewTradeBar(mTradeBarNumber, mModel.Ticker.Timestamp)
            End If
        End If
        sPrevPosition = pPosition
    End Sub

```

### Erroneous output
```C#
        void IStrategyHostController.NotifyPosition(int pPosition)
        {
            ;
#error Cannot convert LocalDeclarationStatementSyntax - see comment for details
            /* Cannot convert LocalDeclarationStatementSyntax, System.NotSupportedException: StaticKeyword not supported!
               at ICSharpCode.CodeConverter.CSharp.SyntaxKindExtensions.ConvertToken(SyntaxKind t, TokenContext context)
               at ICSharpCode.CodeConverter.CSharp.CommonConversions.ConvertModifier(SyntaxToken m, TokenContext context)
               at ICSharpCode.CodeConverter.CSharp.CommonConversions.<ConvertModifiersCore>d__43.MoveNext()
               at System.Linq.Enumerable.<ConcatIterator>d__59`1.MoveNext()
               at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
               at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
               at System.Linq.OrderedEnumerable`1.<GetEnumerator>d__1.MoveNext()
               at Microsoft.CodeAnalysis.SyntaxTokenList.CreateNode(IEnumerable`1 tokens)
[...]

PR #805, opened 2022-01-05
Implement conversion of VB Static variables (#623)

Noah1989:
### Problem
https://github.com/icsharpcode/CodeConverter/issues/623 VB -> C#: Static variables in VB are not converted to C#

### Solution
This uses PerScopeState to keep track of VB static variables. They are then injected as fields above the method in DeclarationNodeVisitor.cs.  The name of the field is based on the method name and the original variable name. The existing name generation scheme is used to avoid identical names, making sure each overload gets its own field. The mechanism for replacing names, which is already used for additional locals, has been extended to be able to replace the name of the static variable with the generated field name.

This does not preserve the order of line comments, since any comment following an initializer for the VB static variables is moved along to the field. I think this is actually useful, therefore I added  `hasLineCommentConversionIssue: true` to the tests.


GrahamTheCoder:
At a first glance this looks great! I'll hopefully have a bit more time to look at it carefully tomorrow for any gotchas.

> ...does not preserve the order of line comments...

Yep, I agree it's correct to move the comment with it. You won't need the cover-all `hasLineCommentConversionIssue` if you just put your own comment at the end of the source line saying `'comment moved to field declaration`, and update the output correspondingly - lines which already have comments don't get tested by the automated ordering check. I realise this behaviour isn't exactly discoverable, perhaps I should make the test failure message explain what I've just said.

CodeConverter/CSharp/PerScopeState.cs
```diff
@@ -102,18 +113,41 @@ public async Task<SyntaxList<StatementSyntax>> CreateLocalsAsync(VBasic.VisualBa
...
+            var newNames = fieldInfo.ToDictionary(f => f.OriginalVariableName, f =>
+                NameGenerator.GetUniqueVariableNameInScope(semanticModel, generatedNames, vbNode, f.FieldName)
+            );
+            foreach (var field in fieldInfo) {
+                var decl = CommonConversions.CreateVariableDeclarationAndAssignment(newNames[field.OriginalVariableName],
+                    field.Initializer, field.Type);
+                var modifiers = new List<SyntaxToken> { CS.SyntaxFactory.Token(SyntaxKind.PrivateKeyword) };
+                if (field.IsStatic) {
+                    modifiers.Add(CS.SyntaxFactory.Token(SyntaxKind.StaticKeyword));
+                }
```
GrahamTheCoder, line 131: Would be good to add a test for the case where the containing type is a module - I think you also need to check here for that case and make the field static

GrahamTheCoder (approved):
Looks perfect other than the very minor things I've mentioned. Ideally update Changelog.md too, thanks!

GrahamTheCoder:
I'll make the tweaks mentioned on master

PR #810, opened 2022-01-11
Fix NRE when a VB static variable has no initializer

Noah1989:
[See comment on #623](https://github.com/icsharpcode/CodeConverter/issues/623#issuecomment-1009917188)

### Problem
When there is no initializer, and also no default initializer is generated because the value is always assigned before the first read, an NRE is thrown because the field generation assumes there always is an initializer.

### Solution
* The field is now generated without an initializer in that case.
* Alternatively, the field could be entirely omitted and the VB Static keyword ignored, because the value is always overwritten anyways, but this way it is closer to the original VB code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### VB -> C#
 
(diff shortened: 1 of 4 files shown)


GrahamTheCoder:
Thanks!

PR #815, opened 2022-01-25
Use new XElement(...) instead of XElement.Parse(...)  for conversions of XML expressions from VB to C#

### LINKED ISSUE #253: VB -> C#: XmlElement conversion not implemented for non-literals
### Input code
```vbnet
Dim element As XElement = <SetAttribute XPath=<%= entry.Code %>></SetAttribute>
```

### Erroneous output
```C#
XElement element = XElement.Parse("<SetAttribute XPath=<%= entry.Code %>></SetAttribute>");
```

### Expected output
```C#
XElement element = new XElement("SetAttribute", new XAttribute("XPath", entry.Code));
```

### Details
Product in use: e.g. web converter

Version in use: newest

Any other relevant information to the issue, or your interest in contributing a fix.

Noah1989:
Fixes #253

### Problem

* The current implementation does not handle embedded expressions well

Also, there are some other minor problems surrounding XML:

* XDocument is assigned using XElement instead (leading to compile error)
* Self-Closing tags are converted incorrectly
* Conditional access (?.) on XML expression throws ArgumentNullException during conversion

### Solution

* Use a per-member visiting approach to construct an actual expression instead of calling XElement.Parse() at runtime (I think that is much closer to what VB actually does)
* Use XDocument constructor in VisitXmlDocument
* Implement VisitXmlEmptyElement to correctly convert self-closing tags
* Use MemberBindingExpression in VisitXmlMemberAccessExpression when node.Base is null (such is the case for conditional access)

The new approach may leave some of the more obscure XML constructs uncovered (falling back to DefaultVisit and thowing a not implemented error) but I think any code that uses those would not convert under the current code either.

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -78,7 +78,7 @@ namespace ICSharpCode.CodeConverter.CSharp
             var convertMethods = convertType.GetMembers().Where(m =>
                 m.Name.StartsWith("To", StringComparison.Ordinal) && m.GetParameters().Length == 1);
             var methodsByType = convertMethods
-                .GroupBy(m => new {ReturnType = m.GetReturnType(), Name = $"{ConvertType.FullName}.{m.Name}"})
+                .GroupBy(m => new { ReturnType = m.GetReturnType(), Name = $"{ConvertType.FullName}.{m.Name}" })
                 .ToDictionary(m => m.Key.ReturnType, m => m.Key.Name);
             return methodsByType;
         }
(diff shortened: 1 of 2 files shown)


CodeConverter/CSharp/ExpressionNodeVisitor.cs
```diff
@@ -97,17 +97,45 @@ public override async Task<CSharpSyntaxNode> DefaultVisit(SyntaxNode node)
...
+                SyntaxFactory.Argument(CommonConversions.Literal(node.Name.ToString())).Yield()
+                .Concat(await node.Attributes.SelectAsync(async attribute => SyntaxFactory.Argument(await attribute.AcceptAsync<ExpressionSyntax>(TriviaConvertingExpressionVisitor))))
+                );
+            return SyntaxFactory.ObjectCreationExpression(SyntaxFactory.IdentifierName("XElement")).WithArgumentList(SyntaxFactory.ArgumentList(arguments));
         }
 
+        public override async Task<CSharpSyntaxNode> VisitXmlAttribute(VBasic.Syntax.XmlAttributeSyntax node)
+        {
+            var arguments = SyntaxFactory.SeparatedList(
+                SyntaxFactory.Argument(CommonConversions.Literal(node.Name.ToString())).Yield()
```
GrahamTheCoder, line 133: It's usually preferable to specify which exact string you want - in this library calling ToString can have surprising effects for various node types or for various escaped values. As far as I know, all xml tag names are valid in a C# literal with no extra escaping though so I think this will probably be fine.

GrahamTheCoder:
Looks like a clear improvement to me, much appreciated!

Issue #848, opened 2022-03-10
Add timestamp for start and finish of project conversion


jrmoreno1:
Converting a large project takes a lot of time and may need to be done multiple times (changing things in the unconverted project to make the conversion work, reconverting an active project multiple times before being ready to commit to new language) 

It would be great if the "Converting MyVbProject ..." message said "Begin converting MyVbProject at hh:mm..."

GrahamTheCoder:
Closing as fixed - thanks!

PR #849, opened 2022-03-10
Issue #848, Modify starting project conversion message in SolutionCon…

jrmoreno1:
Modify starting project conversion message in SolutionConverter, add finished converting message to ConvertProject in ProjectConversion.

#848 

### Problem
It's nice to know how long the conversion has been running and when it finally finishes.

### Solution
It doesn't look like the output messages are being tested, so am not attempting to add a unit test for this change.


GrahamTheCoder (approved):
Good change, thanks (I'll merge once the build has run assuming it succeeds)!

Issue #836, opened 2022-02-13
XML namespace imports are not supported


Noah1989:
Example taken from the [Documentation](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/imports-statement-xml-namespace#example-1):

```vbnet
' Place Imports statements at the top of your program.  
Imports <xmlns="http://DefaultNamespace">
Imports <xmlns:ns="http://NewNamespace">

Module Module1

  Sub Main()
    ' Create element by using the default global XML namespace. 
    Dim inner = <innerElement/>

    ' Create element by using both the default global XML namespace
    ' and the namespace identified with the "ns" prefix.
    Dim outer = <ns:outer>
                  <ns:innerElement></ns:innerElement>
                  <siblingElement></siblingElement>
                  <%= inner %>
                </ns:outer>

    ' Display element to see its final form. 
    Console.WriteLine(outer)
  End Sub

End Module
```

Should produce an XElement that prints like this:

```xml
<ns:outer xmlns="http://DefaultNamespace"
          xmlns:ns="http://NewNamespace">
  <ns:innerElement></ns:innerElement>
  <siblingElement></siblingElement>
  <innerElement />
</ns:outer>
```

I tried reproducing that manually in C# and got the following code: 

```csharp
using System;
using System.Xml.Linq;

public partial class Program
{
    public static void Main()
    {
        var inner = new XElement("{http://DefaultNamespace}innerElement");
        var outer = new XElement("{http://NewNamespace}outer", 
                                 new XAttribute("xmlns", "http://DefaultNamespace"), 
                                 new XAttribute(XNamespace.Xmlns + "ns", "http://NewNamespace"), 
                                 new XElement("{http://NewNamespace}innerElement", ""), 
                                 new XElement("{http://DefaultNamespace}siblingElement", ""), 
                                 inner);
        Console.WriteLine(outer);
    }
}
```

The only weird thing about that are the two `xmlns` attributes. They are only present on the outermost XElement. But how do we know which one is the outermost? Indeed experimenting with VB.NET reveals, that whenever an XElement is created, the compiler adds code that "steals" the attributes from the inner elements. This can be verified by adding the following code to the example: 

```vbnet
		Dim wtf = <foobar><%= outer %></foobar>			
		Console.WriteLine(outer)
```

[...]

PR #852, opened 2022-03-18
Implemet conversion of XML namespace imports

Noah1989:
[Link to issue(s) this covers](https://github.com/icsharpcode/CodeConverter/issues/836)

This PR implements the approach suggested by me in that issue. I decided to keep it at one helper class per file (if needed) to keep the scope where imports apply the same as in VB.

### Problem
[XML import statements](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/imports-statement-xml-namespace#example-1) are not converted.

### Solution
* This solution tries to mimic the behavior of VB XML namespace imports.
* If any XML namespace imports are present, a helper class is generated per file, which contains the namespace definitions. This is in line with how they apply in VB. This also keeps them at the top of the file and allows for changing them easily after conversion.
* The document name is included in the generated helper class name in order to avoid duplicate names while staying deterministic. A using directive is generated to provide a shorter (common) alias.
* When converting XML names with prefixes, the namespace definitions from the helper class are used.
* The helper class is also used to manage xmlns attributes. Whenever a XContainer is created, they are added, and any dupicate attributes from the child elements are removed. This is in line with what the helper code generated by VB does.


GrahamTheCoder:
I'm recovering from quite a bad bout of covid at the moment so don't have the energy to fully understand and give this the in depth review I was hoping to. But from just glancing over it, it's very clearly a big improvement, and looks like good quality code. Thanks a lot for your efforts!

CodeConverter/CSharp/DeclarationNodeVisitor.cs
```diff
@@ -1607,5 +1612,34 @@ public override async Task<CSharpSyntaxNode> VisitTypeConstraint(VBSyntax.TypeCo
...
+                // namespace alias
+                return SyntaxFactory.IdentifierName(node.LocalName.ValueText);
+            } else {
+                // Having this case in VB would cause error BC31187: Namespace declaration must start with 'xmlns'
+                throw new NotImplementedException($"Cannot convert non-xmlns attribute in XML namespace import").WithNodeInformation(node);
+            }
+        }
+
+        public override async Task<CSharpSyntaxNode> VisitXmlString(VBSyntax.XmlStringSyntax node) =>
+            CommonConversions.Literal(node.TextTokens.Aggregate("", (a, b) => a + LiteralConversions.EscapeVerbatimQuotes(b.Text)));
```
GrahamTheCoder, line 1643: Not sure of the likely inputs here myself, but if there could be quite a few strings and especially if they could be big, consider StringBuilder

Noah1989: Agreed. I guess String.Join could also work, right?

GrahamTheCoder: Yep, that'd probably be slightly more readable too actually 👍

Issue #854, opened 2022-03-25
VB -> C#: Handles with differing case is not converted


jrmoreno1:
### VB.Net input code
```vbnet
Public Class VBIsCaseInsensitive
    Inherits System.Web.UI.Page

    Private Sub btnOk_Click(ByVal sender As Object, ByVal e As System.EventArgs) Handles btnOK.Click
    End Sub
End Class

Partial Public Class VBIsCaseInsensitive
    Protected WithEvents btnOk As Global.System.Web.UI.WebControls.Button
End Class

```

### Erroneous output
```C#
class _failedMemberConversionMarker1
{
}
#error Cannot convert ClassBlockSyntax - see comment for details
/* Cannot convert ClassBlockSyntax, CONVERSION ERROR: An item with the same key has already been added. Key: btnOk in 'Public Class VBIsCaseInsens...' at character 203
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey](TSource[] source, Func`2 keySelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)
   at ICSharpCode.CodeConverter.CSharp.HandledEventsAnalysis..ctor(CommonConversions commonConversions, INamedTypeSymbol type, IEnumerable`1 csharpEventContainerToHandlingMethods) in D:\GitWorkspace\CodeConverter\CodeConverter\CSharp\HandledEventsAnalysis.cs:line 35
   at ICSharpCode.CodeConverter.CSharp.HandledEventsAnalyzer.AnalyzeAsync() in D:\GitWorkspace\CodeConverter\CodeConverter\CSharp\HandledEventsAnalyzer.cs:line 58
   at ICSharpCode.CodeConverter.CSharp.DeclarationNodeVisitor.GetMethodWithHandlesAsync(TypeBlockSyntax parentType, IMethodSymbol designerGeneratedInitializeComponentOrNull) in D:\GitWorkspace\CodeConverter\CodeConverter\CSharp\DeclarationNodeVisitor.cs:line 637
[...]

PR #855, opened 2022-03-26
Issue #854: Fix conversion error when Handle clause property differs by case from actual property

jrmoreno1:
Link to issue(s) this covers
#854 

### Problem
When the property declaration and the property reference in the Handles clause, differ by case, the conversion fails trying to create two event containers.

### Solution
* Pass the semantic model to CreateEventContainer so it can be used to get the name from the property declaration
* Change is relatively minor, all unit test continue to pass
* Added new unit test, unit test fails without change, passes after change.
* Copied "hasLineCommentConversionIssue: true /*Fields re-ordered*/);" from another test, not sure what that actually means.


GrahamTheCoder:
Looks good, thanks for that! I've removed `hasLineCommentConversionIssue: true /Fields re-ordered/);`. I'll add a doc comment for it. It's almost never needed. The test passed without it, indicating there isn't a line comment conversion issue, so we don't need to say there is (and thus prevent automatically testing comment conversion).

PR #859, opened 2022-04-03
Add post-assignment to hoisted by-ref argument

### LINKED ISSUE #856: VB -> C#: Double.TryParse into Decimal target results in silently altered behaviour
### VB.Net input code
```vbnet
Module Module1

    Sub Main()
        Dim decimalTarget As Decimal
        Double.TryParse("123", decimalTarget)
    End Sub

End Module
```

### Erroneous output
```C#
namespace ConsoleApp1
{
    static class Module1
    {
        public static void Main()
        {
            var decimalTarget = default(decimal);
            double argresult = (double)decimalTarget;
            double.TryParse("123", out argresult);
        }
    }
}
```

### Expected output
```C#
namespace ConsoleApp1
{
    static class Module1
    {
        public static void Main()
        {
                decimal decimalTarget;
                double.TryParse("123", out var doubleTarget);
                decimalTarget = Convert.ToDecimal(doubleTarget);
        }
    }
}
```

or (leaves an obvious error to resolve)

```C#
namespace ConsoleApp1
{
    static class Module1
    {
        public static void Main()
        {
                decimal decimalTarget;
                double.TryParse("123", out decimalTarget);
        }
    }
}
```

### Details
* Product in use: e.g. VS extension. Website produces approximately the 2nd expected output.
* Version in use: 8.4.7.0
* Did you see it working in a previous version, which? No
* Input is a VB Console project targetting .NET Framework 4.7.2
* Producing the 2nd expected output, which the website version does and is a compiler error, is fine because it would be noticed after the conversion. The current output appears to be a successful conversion with no errors, but silently drops the parsed value resulting in behaviour change.

jconnop:
Add post-assignment to hoisted by-ref argument when argument is an identifer, member access, or array access. Fixes #856

Link to issue(s) this covers
https://github.com/icsharpcode/CodeConverter/issues/856

### Problem
https://github.com/icsharpcode/CodeConverter/issues/856

### Solution
* Return PreAndPostAssignment when expression is a candidate for inline passing if not for the type mismatch.
* All other cases (such as when expression is NothingLiteralSyntax, ObjectCreationSyntax, etc) still just return PreAssignment as before.

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -1662,14 +1662,19 @@ namespace ICSharpCode.CodeConverter.CSharp
                     return propertySymbol.IsReadOnly ? RefConversion.PreAssigment : RefConversion.PreAndPostAssignment;
                 }
 
-                var typeInfo = _semanticModel.GetTypeInfo(expression);
-                bool isTypeMismatch = typeInfo.Type == null || !typeInfo.Type.Equals(typeInfo.ConvertedType);
-
-                if (isTypeMismatch || DeclaredInUsing(symbolInfo)) return RefConversion.PreAssigment;
+                if (DeclaredInUsing(symbolInfo)) return RefConversion.PreAssigment;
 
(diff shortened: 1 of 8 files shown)


GrahamTheCoder:
Awesome thanks!

Issue #866, opened 2022-04-05
Drop support for VS2017 - enable support for newer language features


GrahamTheCoder:
I've already updated the readme to update the intended minimum supported version (VS2019).

I believe that as of March 2022, VS2017 is out of "Mainstream support". Either way, continuing to support it would be a significant burden.

The reason that the CodeConverter project is pinned at 2.8.2 was to support VS2017  - it binds to whichever version Visual Studio loads. For the command line and web, it's fine to bundle later versions, which may fix bugs and improve handling of some syntax types.

To update it we'll need to:
* Remove 2017 from the vsixmanifest (by increasing the required version there)
* Update all dependencies with a 2.8.2 version to the latest version supported by the latest version of VS2019 (latest 3.x)
* Update all dependencies that look Visual Studio related and start with version 15, to the latest version 16
* Update the dependencies in this horrible hack to the versions available in a minimal install of VS2019: https://github.com/icsharpcode/CodeConverter/blob/master/Vsix/Vsix.csproj#L80-L99
  * See if we can (or need to) remove this horribe hack. It was required because VS2017 was missing these common assemblies
* Update readmes and changelog to link to the last working VS2017 version clearly (ideally add a clear message in that version that reminds them there are more features and bugfixes if they use vLatest).
* Test with VS2019 and VS2022

christophwille:
I'd suggest an even more radical approach - move to 2022 entirely. Because if you need the CodeConverter, you can always install the Community edition to do the conversion.

Also, you might want to look into https://github.com/VsixCommunity/Community.VisualStudio.Toolkit to replat the extension onto.

Yozer:
From my experience, I can tell that people tend to convert VB->C# code piece by piece, and having two VS installations can be trouble (opening large solution in another VS instance just to convert code, disk space etc).
I'm not sure if forcing people to install VS2022 is a good idea (are there any benefits for us?).
If we could use abstractions that ```Community.VisualStudio.Toolkit``` gives then switching later from VS2019 to VS2022 shouldn't be that painful.

Do we have any statistics about VS versions that people use when installing extensions?
https://marketplace.visualstudio.com/items?itemName=SharpDevelopTeam.CodeConverter

GrahamTheCoder:
Always worth throwing in radical suggestions! In this case, I don't think it makes sense for reaching the most users, and there isn't a major benefit to it from a development standpoint that balances the downside.

[...]

PR #870, opened 2022-04-09
Vs2019

GrahamTheCoder:
Closes https://github.com/icsharpcode/CodeConverter/issues/866

* [x] Look into VB standalone statement tests
* [x] Check what's happening in SpecialConversionTests with clashing identifiers
* [x] Release master (8.5.0) before this is merged (which has a message encouraging upgrading VS)

### DIFF
--- a/CommandLine/CodeConv.NetFramework/CodeConv.NetFramework.csproj
+++ b/CommandLine/CodeConv.NetFramework/CodeConv.NetFramework.csproj
@@ -10,13 +10,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="5.0.0" />
     <PackageReference Include="Microsoft.Build.Locator" Version="1.2.6" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.4.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="3.4.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.4.0" />
-    <PackageReference Include="Microsoft.VisualStudio.Threading" Version="16.6.13" />
-    <PackageReference Include="Microsoft.Win32.Registry" Version="4.7.0" />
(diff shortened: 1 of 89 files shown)


christophwille:
How about a "legacy" gh release linked from README? Easily discoverable and side loading always works with VS.

GrahamTheCoder:
Yeah, I think let's just do that, and make it as clear as possible on the readme, and extension page where to get an old version so that people don't get confused and/or raise lots of issues about it.

CommandLine/CodeConv.NetFramework/CodeConv.NetFramework.csproj
```diff
@@ -10,13 +10,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="5.0.0" />
```
Yozer, line 13: Func.csproj or CodeConv.csproj are using 6.0.0.

Tests/CSharp/ExpressionTests/ExpressionTests.cs
```diff
@@ -1061,7 +1061,7 @@ private partial class Rec
 
     private Rec TestMethod(string str)
     {
-        int length = str?.Length ?? -1;
+        int length = (str?.Length) ?? -1;
```
Yozer, line 1064: New Roslyn doesn't perform some simplifications?

GrahamTheCoder: Perhaps the default options for simplification changed and this needs explicitly enabling in some way now?
I noticed that one of the options previously used is deprecated too.

CommandLine/CodeConv.NetFramework/CodeConv.NetFramework.csproj
GrahamTheCoder: Ah yes, I think the most correct thing is actually to downgrade CodeConv to 5.0.0 since that's what ships with the latest VS2019

Tests/CSharp/ExpressionTests/ExpressionTests.cs
GrahamTheCoder: I wonder if some of the simplifications were abandoned when trying to make them handle the "target-typed new", and "target-typed conditional expressions"

GrahamTheCoder: I noticed in the renamer that the logic for locating symbols from a different compilation has got worse, I wonder if that's influencing any of the code around simplification

Issue #358, opened 2019-08-15
VB -> C#: String concatenation whitespace trivia lost


mrmonday:
Whitespace trivia gets lost when converting VB to C#

### Input code

```vbnet
Class Program
    Public Shared Sub Main(ByVal args As String())
        Dim x = "a" & "b" &
                "c" & "d" &
                "e" & "f" &
                "g" & "h" &
                "i" & "j" &
                "k" & "l" &
                "m" & "n" &
                "o" & "p"
    End Sub
End Class
```

### Erroneous output
```c#
namespace ConsoleApp4
{
    class Program
    {
        public static void Main(string[] args)
        {
            var x = "a" + "b" + "c" + "d" + "e" + "f" + "g" + "h" + "i" + "j" + "k" + "l" + "m" + "n" + "o" + "p";
        }
    }
}
```

### Expected output

(formatted manually, so the whitespace is probably wrong, despite being visually correct...)

```c#
namespace ConsoleApp4
{
    class Program
    {
        public static void Main(string[] args)
        {
            var x = "a" + "b" +
                    "c" + "d" +
                    "e" + "f" +
                    "g" + "h" +
                    "i" + "j" +
                    "k" + "l" +
                    "m" + "n" +
                    "o" + "p";
        }
    }
}
```

### Details

Version in use: 7.0.0.0

GrahamTheCoder:
At the moment I think whitespace trivia is dropped within expressions in general. Or possibly just when there isn't an accompanying comment.

With the simplifier and formatter running at the end, even if it was converted, it would still often change I think.

I'm reluctant to make any changes to the trivia code until its been tidied up a bit. See #74

Whitespace conversion is definitely at the very bottom of my personal list of priorities, but if anyone's keen I'm happy to help with PRs as always 😊

GrahamTheCoder:
The tidying has now occurred. This definitely makes most sense within expressions. #448 may override some of the formatting anyway

GrahamTheCoder:
Now that the simpler case of between-statement newlines is covered, I think it's more plausible for someone to approach this task. The main tasks here are:
* Figure out how to get the formatter to give reasonable indentation (unlike https://github.com/icsharpcode/CodeConverter/pull/880#discussion_r855317472). It may be just that converted whitespace and generated whitespace are both being converted and hence throwing everything off. "Elastic" trivia may also be key to this.
[...]

PR #880, opened 2022-04-21
Issue 879

### LINKED ISSUE #879: Preserve blank lines during VB>C#
**Is your feature request related to a problem? Please describe.**
When I use this tool to convert VB.NET to C#, it does a great job except it removes almost all of my blank lines.  Having some whitespace between certain groups of code is important to help the developer to easily see logical separation points between lines of code.

Original code:
```
<Display(Name:="Reinsurance Year")> _
Public SelectedReinsuranceYear As Int16

<Display(Name:="Record Type")> _
Public SelectedRecordType As String

<Display(Name:="Release Date")> _
Public ReleaseDate As Nullable(Of Date)
```

gets converted to:
```
[Display(Name = "Reinsurance Year")]
public short SelectedReinsuranceYear;
[Display(Name = "Record Type")]
public string SelectedRecordType;
[Display(Name = "Release Date")]
public DateTime? ReleaseDate;
```

Another example:
```
Dim listOfThings As New List(Of String)

With listOfThings
    .Add("section 1 - line 1")
    .Add("section 1 - line 2")
    .Add("section 1 - line 3")

    .Add("section 2 - line 1")
    .Add("section 2 - line 2")
    .Add("section 2 - line 3")
End With
```
gets converted to:
```
List<string> listOfThings = new List<string>();
listOfThings.Add("section 1 - line 1")
listOfThings.Add("section 1 - line 2")
listOfThings.Add("section 1 - line 3")
listOfThings.Add("section 2 - line 1")
listOfThings.Add("section 2 - line 2")
listOfThings.Add("section 2 - line 3")
```
but it should keep those line breaks that I put there on purpose for readability:
```
List<string> listOfThings = new List<string>();

listOfThings.Add("section 1 - line 1")
listOfThings.Add("section 1 - line 2")
listOfThings.Add("section 1 - line 3")

listOfThings.Add("section 2 - line 1")
listOfThings.Add("section 2 - line 2")
listOfThings.Add("section 2 - line 3")
```

**Describe the solution you'd like**
Never remove blank lines while converting languages, even if there is more than one in a row.

**Describe alternatives you've considered**
You could make that behavior configurable so anyone who wants to have all their blank lines purged can enable that setting.

**Additional context**
Code Converter (VB - C#) version 8.5.00
Visual Studio 2022 version 17.1.4

GrahamTheCoder:
Fixes #879 

### Problem
New lines not converted. This PR focuses on between-statement newlines (not within-expression newlines)

### Solution
Because the shape of the syntax tree in the languages isn't the same, it's quite hard to pick the right element to move newlines between.

In general, the system relies on the Roslyn formatter to add in some sensible line breaks, then attempts to map any line comments so that they appear at the end of an existing line between the converted versions of elements they were between in the source.

This PR includes newlines in that system too. However, this means newlines sometimes get saved up and shuffled much further than they should. The two remedies are:
* Add in more line breaks by converting some trivia prior to formatting (I've tried this on XmlElement)
* Stop saving up newlines. Code comments contain more information and hence are worth placing even in the wrong place, newlines only have any meaning if they are mapped precisely. (I haven't tried to apply this yet).


Tests/CSharp/ExpressionTests/XmlExpressionTests.cs
```diff
@@ -114,9 +114,36 @@ internal partial class TestClass
...
+                                       select new XElement(""div"",
+                 new XElement(""h1"", book.Elements(""Title"").Value),
+                 new XElement(""h3"", ""By "" + book.Elements(""Author"").Value),
+                 new XElement(""h3"", ""Price = "" + book.Elements(""Price"").Value),
+
+
+                                                  new XElement(""h2"", ""Description""), TransformDescription((string)book.Elements(""Description"").ElementAtOrDefault(0)), new XElement(""hr"")
+                 )
+                    )
+                );
```
GrahamTheCoder, line 146: This is reasonably horrendous since the indentation is wrong, but the newlines from the source are preserved in a way that means an autoformatter can do a good job on the output while being consistent with the original source.
I'll leave any further work on this to be covered by issue: https://github.com/icsharpcode/CodeConverter/issues/358

PR #891, opened 2022-05-11
Bugfixes for XML Import Helper code

Noah1989:
These changes to the boilerplate template fix run-time errors in some special cases:

- Generating an XDocument with imported XML namespaces: This used to throw an Exception because the xmlns attributes cannot be added to the Document itself. Add to the root instead.
- Make Apply idempotent by including the main element in the scan for duplicate attributes, so that no "Duplicate Attribute" error can occur, even when Apply happens to be called twice on the same element.
- Add readonly attribute to make linter happy

### DIFF
--- a/CodeConverter/CSharp/XmlImportContext.cs
+++ b/CodeConverter/CSharp/XmlImportContext.cs
@@ -36,7 +36,7 @@ internal class XmlImportContext
 
         var xAttributeList = SyntaxFactory.FieldDeclaration(
             SyntaxFactory.List<AttributeListSyntax>(),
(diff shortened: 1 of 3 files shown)


GrahamTheCoder:
Looks good! Is it possible to add the issues you talked about as individual tests? (from what I see they don't seem to be covered)

Noah1989:
These issues only occur when running the generated code. Is there a preferred way to do that in a test?

GrahamTheCoder:
Ah yes, good question. You'll want to use a self verifying test mentioned here https://github.com/icsharpcode/CodeConverter/blob/master/Tests/Readme.md#types
Basically write a test case that passes in VB but would fail/throw if converted incorrectly.
Let me know if you can't see how to get started from the existing ones

Noah1989:
I've added the self-verifying tests for those issues and verified that they would fail If the change were reverted.

Yozer (approved):

Issue #893, opened 2022-05-12
VB -> C#: wrong conversion of generic function with CType


KunzeAndreas:
### VB.Net input code
```vbnet
Private Shared Sub GenericCast()
    Assert.That(GenericFunctionWithCast(Of Integer)() = 1)
    Assert.That(GenericFunctionWithCast(Of Short)() = 1)
End Sub

Private Shared Function GenericFunctionWithCast(Of T)() As T
    Const result = 1
    Dim resultObj As Object = result
    Return CType(resultObj, T)
End Function
```

### Erroneous output
```C#
private static T GenericFunctionWithCast<T>()
{
    const int result = 1;
    object resultObj = result;

   return (T)resultObj;     // run time error when T is short
}
```

### Expected output
```C#
return Microsoft.VisualBasic.CompilerServices.Conversions.ToGenericParameter<T>(resultObj);
```

### Details
* Product in use: VS extension 
* Version in use: 9.0.0.0
* Did you see it working in a previous version, which? no

PR #899, opened 2022-05-24
fix for VB -> C#: wrong conversion of generic function with CType #893

Noah1989:
[ VB -> C#: wrong conversion of generic function with CType #893](https://github.com/icsharpcode/CodeConverter/issues/893)

### Problem
When casting to a generic type in VB, the converted C# code can fail at runtime where VB doesn't.
`return (T)resultObj;     // run time error when T is short`

### Solution
* Use Call to `Microsoft.VisualBasic.CompilerServices.ToGenericParameter<>` instead of simple cast


Yozer (approved):
Thanks! Looks good.
I updated the changelog and added two tests:
- cast without using CType
- cast that exists in C# so Conversions.ToGenericParameter isn't needed (and it's green!)

Issue #897, opened 2022-05-16
VB -> C#: variable is initialized inside loop


KunzeAndreas:
### VB.Net input code
```vbnet
Private Shared Sub DefaultInitialization()
    For i = 1 To 2
        Dim b As Boolean
        If i = 1 Then
            Assert.That(Not b)
        Else
            Assert.That(b)
        End If

       b = True
       Assert.That(b)
   Next
 End Sub
```

### Erroneous output
```C#
private static void DefaultInitialization()
{
    for (int i = 1; i <= 2; i++)
    {
        var b = default(bool);
        if (i == 1)
        {
            Assert.That(!b);
        }
        else
        {
            Assert.That(b);
        }

        b = true;
        Assert.That(b);
    }
}
```

### Expected output
```C#
private static void DefaultInitialization()
{
    var b = default(bool);  // define and initialize outside loop or do not initialize (with compile error CS0165)
    for (int i = 1; i <= 2; i++)
    {
        // bool b;
        if (i == 1)
        {
            Assert.That(!b);
        }
        else
        {
            Assert.That(b);
        }

        b = true;
        Assert.That(b);
    }
}
```

### Details
* Product in use: VS extension 
* Version in use: 9.0.0.0
* Did you see it working in a previous version, which? no

GrahamTheCoder:
VB still manages to surprise me after all these years. If I ever knew about that "feature" of VB (pulling the variable into the outer scope), I've erased it from my mind. I guess it's because VB only allows one loop variable but you might want multiple, and there's no other way to limit the scope to within the loop since blocks don't exist. Whatever the reason, yep it definitely needs fixing in the converter!

GrahamTheCoder:
[...]

PR #902, opened 2022-06-15
Fix Issue #897 (VB -> C#: variable is initialized inside loop)

Noah1989:
[Link to issue this covers](https://github.com/icsharpcode/CodeConverter/issues/897)

### Problem
Uninitialized variables are implicitly set to the default value in VB. The converter needs to add the initialization explicitly. However, in VB the initialization only happens once, even when the variable declaration is inside a loop. Therefore the converted code behaves differently.

### Solution
* Pull default initialized variables out of the loop to ensure they are only initialized once.
* This also applies to nested loops: variables are pulled all the way out to ensure the same behavior as in VB.
* Nested loops in VB can lead to different variables with the same name that end up in the same scope after conversion. These have their names changed.
* [x] At least one test covering the code changed -> I have added a simple case as described in the issue, as well as a complicated one with all loop types and nested clashing variable names. Both cases have a self-verifying test as well as a normal one.

### DIFF
--- a/CodeConverter/CSharp/HoistedDefaultInitializedLoopVariable.cs
+++ b/CodeConverter/CSharp/HoistedDefaultInitializedLoopVariable.cs
@@ -1,0 +1,24 @@
+﻿using System.Diagnostics;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace ICSharpCode.CodeConverter.CSharp;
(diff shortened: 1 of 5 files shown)


GrahamTheCoder:
Looks great, nice catch on the case where the same variable name is pulled out of two loops!

Issue #895, opened 2022-05-12
VB -> C#: missing where and wrong select in linq with group clause


KunzeAndreas:
### VB.Net input code
```vbnet
Private Shared Sub LinqWithGroup()
    Dim numbers = New List(Of Integer) From {1, 2, 3, 4, 4}
    Dim duplicates = From x In numbers
                     Group By x Into Group
                     Where Group.Count > 1
    Assert.That(duplicates.Count = 1)
End Sub
```

### Erroneous output
```C#
var duplicates = from x in numbers
                 group x by x into Group
                 select x;
```

### Expected output
```C#
var duplicates = from x in numbers
                     group x by x 
                     into Group
                     where Group.Count() > 1
                     select Group;
```

### Details
* Product in use: VS extension 
* Version in use: 9.0.0.0
* Did you see it working in a previous version, which? no

GrahamTheCoder:
Thanks for the report. The area of code for query syntax is terrible, so this may not get fixed for some time I'm afraid

Noah1989:
I'll have a look at what I can do about this one :)

PR #903, opened 2022-06-16
Fix Issue #895 (VB -> C#: missing where and wrong select in linq with group clause)

Noah1989:
[Link to issue(s) this covers](https://github.com/icsharpcode/CodeConverter/issues/895)

### Problem
Nested clauses after a group by get dropped, also the wrong default selection is used.

### Solution
* Pass along the nested clauses to CreateGroupByContinuation, so they no longer get lost.
* When building a "group by" continuation, do not use the selection from the nested clause if it is the default selection, select the group itself instead.


GrahamTheCoder:
Nice one! I keep expecting a query to come along that just can't be handled by the current code structure  and prompt a rewrite of the whole area, but it lives to execute another day

Issue #795, opened 2021-11-28
Github actions release flow


GrahamTheCoder:
For maximum consistency and convenience, we could automate some/all of the release process based on pushing a release tag

Examples of releasing based on tag from our other stuff:
https://github.com/icsharpcode/ilspy-vscode/blob/master/.github/workflows/ci.yml#L46
https://github.com/icsharpcode/SharpZipLib/blob/master/.github/workflows/release.yml

Azure static site:
https://github.com/christophwille/github-release-stats

Azure function:
https://github.com/marketplace/actions/azure-functions-action

Artifacts/changelog as github release:
https://github.com/softprops/action-gh-release#%EF%B8%8F-uploading-release-assets

Nuget:
https://stackoverflow.com/a/58025355/1128762

VS marketplace:
https://cezarypiatek.github.io/post/develop-vsextension-with-github-actions/#release-workflow

christophwille:
My preference would be to publish the static Web page to GH Pages, and use Azure Functions as the backend API.

PR #926, opened 2022-07-29
Push to github pages

GrahamTheCoder:
Part of https://github.com/icsharpcode/CodeConverter/issues/795

Irritatingly the action I'm using only runs on ubuntu so it has to be a totally separate job. IIRC we need to build on windows for some legacy msbuild related stuff to work

TODO
* [x] Once this works first time, restrict to master
`if: ${{ github.ref == 'refs/heads/master' }}`

@christophwille You'll need to setup the repo settings something like this:
image

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -78,6 +78,7 @@ jobs:
         path: Func/bin/${{ env.BuildTarget }}/publish/
         
  deploy:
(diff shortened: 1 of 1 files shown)


GrahamTheCoder:
You can see the site root here: https://github.com/icsharpcode/CodeConverter/tree/autoupdated/gh-pages

PR #928, opened 2022-08-03
Pass CancellationToken

christophwille:
As an extension, I'd like to suggest passing ILoggerFactory? into the call chain so we can create logger instances (when non-null) to log errors and information (that can be surfaced in case of the Azure Functions deployment using Application Insights).

### DIFF
--- a/Func/Convert.cs
+++ b/Func/Convert.cs
@@ -1,4 +1,5 @@
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using ICSharpCode.CodeConverter.Web;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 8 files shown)


Func/Startup.cs
```diff
@@ -0,0 +1,22 @@
...
+        public override void Configure(IFunctionsHostBuilder builder)
+        {
+            //builder.Services.AddHttpClient();
+
+            //builder.Services.AddSingleton<IMyService>((s) => {
+            //    return new MyService();
+            //});
+
+            //builder.Services.AddSingleton<ILoggerProvider, MyLoggerProvider>();
+        }
```
GrahamTheCoder, line 20: What is this for?

Func/Convert.cs
```diff
@@ -10,20 +11,31 @@
...
     //
     [FunctionName("Convert")]
 #pragma warning disable VSTHRD200 // Use "Async" suffix for async methods - Name must be "Run" for this to work AFAIK
-    public static async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] HttpRequest req, ILogger log)
+    public async Task<IActionResult> Run(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] HttpRequest req,
+        CancellationToken hostCancellationToken)
 #pragma warning restore VSTHRD200 // Use "Async" suffix for async methods
     {
+        var logger = _loggerFactory.CreateLogger<Convert>();
```
GrahamTheCoder, line 33: Will this violate what the site says about not storimg code in any way?
Exception messages contain fragments of code

Func/Startup.cs
christophwille: This is directly from the sample for how-to add injection to Azure Functions. I kept those lines as comment for "inspiration".

Func/Convert.cs
christophwille: This is not yet logging anything, this is a verification of ILoggerFactory injection. And yes, that is exactly why I didn't sprinkle any of it in CC itself because we need to have the discussion how to make this logging useful to us without violating that promise (even maybe having two paths - because in the extension it might be useful to surface the information to the user via logging whereas we in the Azure Functions case are only interested in the high-level information)

GrahamTheCoder (approved):
If you're happy it works, feel free to merge and either release with 9.03 or afterwards

PR #937, opened 2022-08-05
Short-circuit UI default code

christophwille:


GrahamTheCoder:
I'll try to remember to use something more complex as the basic check things are working from now on.
If this turns out to have little effect on costs let's remove it again for simplicity

christophwille:
Maybe move it to the UI so the UI detects "oh, this is the sample" and does the magic locally without calling out at all.

Issue #982, opened 2023-01-19
VB -> C#: Problem converting long if statement with nullable integers


cysieks:
### VB.Net input code
```vbnet
'Dim newDays As Integer?
'Dim oldDays As Integer?

If (newDays.HasValue AndAlso Not oldDays.HasValue) _
                OrElse (newDays.HasValue AndAlso oldDays.HasValue AndAlso newDays <> oldDays) _
                OrElse (Not newDays.HasValue AndAlso oldDays.HasValue) Then

'Some code
End If
```

### Erroneous output
```C#
if ((bool)((newDays.HasValue && !oldDays.HasValue ? true : newDays.HasValue && oldDays.HasValue ? newDays is var arg1 && oldDays is { } arg2 && arg1.HasValue ? arg1 != arg2 : null : false).GetValueOrDefault() || !newDays.HasValue && oldDays.HasValue))
// Some code


```

### Expected output
```C#
if ((newDays.HasValue && !oldDays.HasValue) || (newDays.HasValue && oldDays.HasValue && newDays != oldDays) || (!newDays.HasValue && oldDays.HasValue))
```

### Details
* Product in use: e.g. both
* Version in use: e.g. 9.1.2
* Did you see it working in a previous version, which? No
* Any other relevant information to the issue, or your interest in contributing a fix.

GrahamTheCoder:
Thanks for this. It looks like the code dealing with nullables in boolean statements could do with some special cases for things like HasValue to tidy them up.

cysieks:
Yes, the same is with e.g. nullable enums. We have a huge solution that needs to be migrated to C#, and this becomes a blocker.
Thank you for the great tool.

GrahamTheCoder:
Just to confirm, have you seen the code producing different results in any case, or is it "just" that code looks terrible?

cysieks:
Hi,

During conversion we have multiple strange results like these:

```
 adminDate = administrationAccess.SelectAdminDate(XXX) 'this is nullable
                
If adminDate Is Nothing OrElse adminDate > chartingTimeAllowanceEnd Then
    adminDate = Clock.Now
End If
```

is converted to:

`if ((bool)(adminDate is null || (chartingTimeAllowanceEnd is var arg2 && adminDate is { } arg1 ? arg1 > arg2 : (bool?)null).GetValueOrDefault()))`

or

```
Dim status As PasswordStatus? = securityManager.GetUserPasswordStatus(userName)

[...]

PR #984, opened 2023-01-29
Issue/982

GrahamTheCoder:
Try to cover some common cases mentioned in #982

It's a tricky area since I don't want to change the number of evaluations of any expression involved that may have side effects
There are a lot more cases which could be tidied up, see https://github.com/icsharpcode/CodeConverter/issues/982#issuecomment-1407665870 but I think I'll merge just these first steps for now

I've only applied improvements to relational operators. It's likely the same patterns (checking for reusable expressions and using flow analysis) could be applied to the special-cased OrElse and AndAlso cases in future. It's just a case of finding time to do it.

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -897,8 +897,10 @@ internal class ExpressionNodeVisitor : VBasic.VisualBasicSyntaxVisitor<Task<CSha
                 return _visualBasicEqualityComparison.GetFullExpressionForVbObjectComparison(lhs, rhs, node.IsKind(VBasic.SyntaxKind.NotEqualsExpression));
         }
 
-        omitConversion |= lhsTypeInfo.Type != null && rhsTypeInfo.Type != null &&
-                          lhsTypeInfo.Type.IsEnumType() && SymbolEqualityComparer.IncludeNullability.Equals(lhsTypeInfo.Type, rhsTypeInfo.Type)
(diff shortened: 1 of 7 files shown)


CodeConverter/CSharp/ExpressionNodeVisitor.cs
```diff
@@ -897,8 +897,10 @@ public override async Task<CSharpSyntaxNode> VisitBinaryExpression(VBasic.Syntax
                 return _visualBasicEqualityComparison.GetFullExpressionForVbObjectComparison(lhs, rhs, node.IsKind(VBasic.SyntaxKind.NotEqualsExpression));
         }
 
-        omitConversion |= lhsTypeInfo.Type != null && rhsTypeInfo.Type != null &&
-                          lhsTypeInfo.Type.IsEnumType() && SymbolEqualityComparer.IncludeNullability.Equals(lhsTypeInfo.Type, rhsTypeInfo.Type)
+        var lhsTypeIgnoringNullable = lhsTypeInfo.Type.GetNullableUnderlyingType() ?? lhsTypeInfo.Type;
+        var rhsTypeIgnoringNullable = rhsTypeInfo.Type.GetNullableUnderlyingType() ?? rhsTypeInfo.Type;
+        omitConversion |= lhsTypeIgnoringNullable != null && rhsTypeIgnoringNullable != null &&
+                          lhsTypeIgnoringNullable.IsEnumType() && SymbolEqualityComparer.Default.Equals(lhsTypeIgnoringNullable, rhsTypeIgnoringNullable)
```
GrahamTheCoder, line 903: This was the Conversions.ToInteger issue. The ConvertedType says `int?` since that's presumably just always how enums are compared under the covers (hence this existing code), but because one side was nullable, this omitConversion option wasn't set

Issue #426, opened 2019-11-15
VB -> C#: OnErrorGoToStatementSyntax and OnErrorResumeNextStatementSyntax


GrahamTheCoder:
**2023-12-10: Yet to find an acceptable general way to handle this since they're all way too messy. Still happy to take specific common examples that could be easily handled**

 * `OnErrorGoToStatementSyntax`, `OnErrorResumeNextStatementSyntax`
   * Many cases can probably be reasonably successfully converted as `try...catch` and then a goto within the catch and an appropriately placed label

### Resume Next Example

```csharp
    int lineNumber = 1;
    try
    {
        line1: lineNumber++;
        //First line of code would be here
        line2: lineNumber++;
        //Second line of code would be here    
    }
    catch
    {
        switch(lineNumber)
        {
            case 1: goto line1;
            case 2: goto line2;
        }
    }
```
Note: Even this messy solution isn't valid because line1 is out of scope of the catch block and could be in a nested scope even within the try block

valeriu33:
I found this issue as well, my project has a lot of "On Error GoTo"s and I'd be grateful if someone implemented this syntax
![image](https://user-images.githubusercontent.com/29232328/75870194-51eb8300-5e13-11ea-9ade-4fe9b5250209.png)

GrahamTheCoder:
On Error Goto is easy enough to convert cleanly into a try catch.

The "Resume Next" is problematic since it forces the code to resume from the next line. Things I've considered:
* Wrapping every single line in a try catch
* Using a counter variable to track which line was reached
* Encasing every statement in a lambda

Perhaps the generated IL will give some further ideas, please post any ideas that come to mind here.

GrahamTheCoder:
The neatest solution I've thought of for `Resume Next` is:

```csharp
for (int i = 1; i <= 2; i++)
{
    try
    {
        switch(i)
        {
            case 1:
             //First line of code would be here
            break;
            case 2:
             //Second line of code would be here
            break;
        }
    }
    catch
    {
    }
}
```

GrahamTheCoder:
Actually this would save us having to mess around with variable scopes
```csharp
    int lineNumber = 1;
    try
    {
        line1: lineNumber++;
        //First line of code would be here
        line2: lineNumber++;
        //Second line of code would be here    
    }
    catch
    {
        switch(lineNumber)
        {
            case 1: goto line1;
            case 2: goto line2;
        }
    }
```

[...]

PR #999, opened 2023-03-26
OnError GoTo

GrahamTheCoder:
Partial (simplest possible case) fix for https://github.com/icsharpcode/CodeConverter/issues/426

Simplest case covered here: A single top level On Error Go To, everything after it could be placed in a try catch and go to the outer label

### May be possible to cover in future

* OnErrorGoTo statement and label in same scope, but not top-level (e.g. nested in an if statment/loop)

* Extend that system to two statements within a scope (e.g. changing the error handler part way through, it would be possible to have a nested try catch block for the second one, and use an filter clause to avoid hitting the outer one

 

### Still no known solution

General solution: The difficulty is that in C# the labels are scoped to the containing block, so it isn't possible to goto a label straight from a catch block. I've therefore created a loop and switch to allow it to re-enter the scope where it's possible to go to the label. I'm still trying to think of a less verbose/ugly way of handling this.

### DIFF
--- a/Tests/CSharp/StatementTests/OnErrorStatementTests.cs
+++ b/Tests/CSharp/StatementTests/OnErrorStatementTests.cs
@@ -15,9 +15,10 @@ Public Function SelfDivisionPossible(x as Integer) As Boolean
         Dim i as Integer = x / x
         Return True
 ErrorHandler:
-    Return False
+    Return Err.Number = 6
 End Function
 End Class", @"using System;
+using Microsoft.VisualBasic; // Install-Package Microsoft.VisualBasic
 
 internal partial class TestClass
 {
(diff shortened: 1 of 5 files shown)


Tests/CSharp/StatementTests/OnErrorStatementTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            ErrorHandler:
+                ;
+
+                Console.Write(0);
+                catchIndex = -1;
+            }
+            catch
+            {
+            }
+        while (catchIndex != -1);
```
GrahamTheCoder, line 57: This is the example output - which is what's discouraging me from further work on this...

Issue #998, opened 2023-03-26
VB -> C#: Handling of nested default-initialized loop variables


philippejer:
### VB.Net input code
```vbnet
Class TestClass
    Private Sub TestMethod(someCondition As Boolean)
        For j = 1 To 2
            If someCondition Then
                Dim b As Boolean
                Console.WriteLine(b)
                b = True
            End If
        Next
    End Sub
End Class
```

### Erroneous output
```C#
using System;
          
internal partial class TestClass
{
    private void TestMethod(bool someCondition)
    {
        var b = default(bool);
        for (int j = 1; j <= 2; j++)
        {
            if (someCondition)
            {
                Console.WriteLine(phb9bd034daca847259b64966be09448f5);
                phb9bd034daca847259b64966be09448f5 = true;
            }
        }
    }
}
```

### Expected output
```C#
using System;
          
internal partial class TestClass
{
    private void TestMethod(bool someCondition)
    {
        var b = default(bool);
        for (int j = 1; j <= 2; j++)
        {
            if (someCondition)
            {
                Console.WriteLine(b);
                b = true;
            }
        }
    }
}
```

### Details
* Product in use: Found in VS extension, reproduced in latest master
* Version in use: VS extension 9.2.0.0
* Did you see it working in a previous version, which? Yes, see below.

~I managed to track down the issue to this PR: https://github.com/icsharpcode/CodeConverter/pull/902~

~It appears the issue is that the `value` variable is declared inside an `If` statement but the conversion seems to handle it as a local variable inside a loop.~

~Explicitly initializing the `value` variable on the VB side yields the expected output (as well as removing any of the control flow statements).~

Edited after working on the issue : 
* Miminized the test case.
* #902 is not a regression, the variable must be pulled ouside the loop even if not directly under the loop statement.
It is simply a bug in the implementation of the fix.

GrahamTheCoder:
Thanks for the report. There's code to replace that temporary generated name - not sure why it's not working here.
Let me know if you are interested in trying to fix it, otherwise hopefully I'll get around to it at some point ☺

philippejer:
I've looked at the code, I'll try to see if I can have a clean fix (given my lack of familiarity with the code 😊)

GrahamTheCoder:
[...]

Issue #1000, opened 2023-04-13
VB -> C#: Handling of default-initialized loop variables in sibling loops


philippejer:
### VB.Net input code
```vbnet
Class TestClass
    Private Sub TestMethod()
        For i = 1 To 2
            Dim b As Boolean
            Console.WriteLine(b)
            b = True
        Next
        For i = 1 To 2
            Dim b As Boolean
            Console.WriteLine(b)
            b = True
        Next
    End Sub
End Class
```

### Erroneous output
```C#
using System;
          
internal partial class TestClass
{
    private void TestMethod()
    {
        var b = default(bool);
        for (int i = 1; i <= 2; i++)
        {
            Console.WriteLine(b);
            b = true;
        }
        var b = default(bool);
        for (int i = 1; i <= 2; i++)
        {
            Console.WriteLine(b);
            b = true;
        }
    }
}
```

### Expected output
```C#
using System;
          
internal partial class TestClass
{
    private void TestMethod()
    {
        var b = default(bool);
        for (int i = 1; i <= 2; i++)
        {
            Console.WriteLine(b);
            b = true;
        }
        var b1 = default(bool);
        for (int i = 1; i <= 2; i++)
        {
            Console.WriteLine(b1);
            b1 = true;
        }
    }
}
```

### Details
* Product in use: Found in VS extension, reproduced in latest master
* Version in use: VS extension 9.2.0.0

Found this issue while working on #998, conflicting variables in sibling loops are not renamed (unless nested under a parent loop).

philippejer:
I have been able to investigate this issue (and #998), I'll create a PR shortly.

PR #1001, opened 2023-04-13
Fix Issue #998 and #1000

philippejer:
### Problem
See PR #902 for the explanation of the root issue. Issues #998 and #1000 are bugs in #902 in two specific cases.

### Solution
The fix for the two issues is in the same location `PerScopeState.CreateLocalsAsync()`
* For #998
Renaming the original variable with a unique ID as we go up the tree should only be done (and is only needed) in the nested loops case.
* For #1000 
Renaming the original variable in case of conflict was only done in the nested loops case, but a conflict can also happen in the non-nested case (see test).
A side effect of the fix is that the renaming occurs only in case of conflict (see modified test for the nested loops case).
* I have tried to merge the nested and non-nested code paths, but currently I don't think it is possible.


GrahamTheCoder:
Thanks for this, looks like a good improvement!

Issue #1003, opened 2023-04-16
VB -> C#: DefineDebug and DefineTrace are not converted


philippejer:
### VB.Net input project file
```xml
<PropertyGroup>
  <DefineDebug>true</DefineDebug>
  <DefineTrace>true</DefineTrace>
</PropertyGroup>
```

### Erroneous output
```xml
<PropertyGroup>
  <DefineDebug>true</DefineDebug>
  <DefineTrace>true</DefineTrace>
</PropertyGroup>
```

### Expected output
```xml
<PropertyGroup>
  <DefineConstants>TRACE;DEBUG</DefineConstants>
</PropertyGroup>
```

### Details
* Product in use: VS extension and run directly from latest master branch
* Version in use: VS extension 9.2.1.0

VB project files typically use `<DefineDebug>` and `<DefineTrace>` to define the `DEBUG` and `TRACE` preprocessor variables instead of `<DefineConstants>` (https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2022).

For some reason, CS project files ignore these, so currently a VB project converted to CS will not have `DEBUG` defined in debug configuration.

Hence `<DefineDebug>` and `<DefineTrace>` should be converted to `<DefineConstants>`, as explained in this comment: https://github.com/icsharpcode/CodeConverter/blob/4ea68bd3f9c117055de993696cafa03c8c2acbc8/CodeConverter/CSharp/VBToCSConversion.cs#L76-L89

philippejer:
I'll create a PR for this issue in a moment.

PR #1004, opened 2023-04-16
Fix issue #1003

philippejer:
See issue #1003

### Problem
For the VB to CS project file conversion, `<DefineDebug>` and `<DefineTrace>` should be converted to `<DefineConstants>`.

### Solution
* I've used `XDocument` to parse and edit the XML (this is already used elsewhere in the codebase, e.g. to parse the resource files).
* I've also done a similar job in the neighborhood for consistency and to simplify the code (especially in the case of the `<DocumentationFile>` conversion).
* The main change in external behavior (aside from fixing the issue) is that the `XDocument` step can cause minor changes in formatting (this can be seen in the updated test files), but I don't think this is an issue (they are new files generated by the converter, not existing files).
* I've done a small change in `MultiFileTestFixture` to ouput the expected file path in case of failure.
* The code is already covered quite well by the tests, I've updated the relevant test cases.

### DIFF
--- a/CodeConverter/CSharp/VBToCSConversion.cs
+++ b/CodeConverter/CSharp/VBToCSConversion.cs
@@ -1,6 +1,7 @@
 ﻿using System.Text;
 using SyntaxKind = Microsoft.CodeAnalysis.VisualBasic.SyntaxKind;
 using System.Text.RegularExpressions;
+using System.Xml.Linq;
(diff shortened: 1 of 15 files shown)


GrahamTheCoder:
Great job, thanks for this, it's much nicer parsing the xml than purely using regexes everywhere

PR #1031, opened 2023-08-25
Fix `VisualBasic.Chr` method recognition

### LINKED ISSUE #1030: VB -> C#: own `Chr` method invocations are lost with terrible consequences
### VB.Net input code
```vbnet
Public Class ConvertChr
    Sub UseChr()
        Dim a As String
        Dim b As String
        b = Chr(2) & a$ & Chr(13) & Chr(3)
    End Sub

    Function Chr(o As Object) As Char
        Return Microsoft.VisualBasic.Chr(o)
    End Function
End Class
```

### Erroneous output
```C#
    public class ConvertChr
    {
        public void UseChr()
        {
            var a = default(string);
            string b;
            b = 2 + a + 13 + 3;
        }

        public char Chr(object o)
        {
            return Microsoft.VisualBasic.Strings.Chr(Conversions.ToInteger(o));
        }
    }
```

### Expected output
```C#
public class ConvertChr
    {
        public void UseChr()
        {
            var a = default(string);
            string b;
            b = Conversions.ToString(Chr(2)) + a + Chr(13) + Chr(3);
        }

        public char Chr(object o)
        {
            return Microsoft.VisualBasic.Strings.Chr(Conversions.ToInteger(o));
        }
    }
```

### Details
* Version in use: 9.2.3.0
* I haven't checked previous versions, so I'm not sure if it ever worked

TymurGubayev:
Fixes #1030

### Problem
Any method called `Chr` was treated as if it was `Microsoft.VisualBasic.Chr`.

### Solution
* Add a test case for custom methods called `Chr` and `ChrW`
* Move the check for `Chr` and `ChrW` in a separate method
* Adjust the check to not have false-positive result
* Also, in `IsCultureInvariant` remove the symbol name check as redundant


GrahamTheCoder:
Thanks for this, and sorry it took me so long to get it merged!

PR #1036, opened 2023-10-05
fix: ClashingMemberRenamer incorrectly renames Enum types

### LINKED ISSUE #1035: VB -> C#: ClashingMemberRenamer incorrectly renames Enum types
### VB.Net input code
```vbnet
Public Enum Column
    Column
    ColumnPosition
End Enum
```

### Erroneous output
```C#
public enum Column
{
    ColumnField,
    ColumnPositionField
}
```

### Expected output
```C#
public enum Column
{
    Column,
    ColumnPosition
}
```

### Details

There is a logic to convert any enum members that are prefixed with the enum name.
I found [this ](https://github.com/icsharpcode/CodeConverter/issues/420#issuecomment-596098221)old issue which mention that this is not allowed in C#:
But I wasn't able to find anything in language specification that would not allow such naming (even if enum member is equal to enum itself).
[Sharplab.io](https://sharplab.io/#v2:EYLgHgbALANALiATgVwHYwCYgNQB8ACATAIwCwAUPsQJwAUqAhgLYCmA9gGa0DCbANsiaoAdLwFCAlBIDcFKnTGCRioQFFhAFTYBlOIgCWqAOa0ps8nIDMAAhapB1lagoBvCtY+P+SmO89PVCgBfIA==)

I found a recommendation but it results in a warning, not error:
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1712

* Product in use: e.g. [icsharpcode.github.io/CodeConverter](https://icsharpcode.github.io/CodeConverter/) / **VS extension** / both
* Version in use: e.g. 5.6.3 or a commit hash https://github.com/icsharpcode/CodeConverter/commit/2ea789bc9c3123c8caa23ea1989bd6c65675354b
* Did you see it working in a previous version, which?
* Any other relevant information to the issue, or your interest in contributing a fix.

Yozer:
### Problem
Fixes #1035 

### Solution

// few tests are failing, will take a look at them later...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * Omit special VB conversions within expression trees [#930](https://github.com/icsharpcode/CodeConverter/issues/930) [#316](https://github.com/icsharpcode/CodeConverter/issues/316)
 * Support CData [#1032](https://github.com/icsharpcode/CodeConverter/issues/1032)
 * Use verbatim strings for strings containing newlines
+* Fix clashing symbol renamer for Enum types [#1035](https://github.com/icsharpcode/CodeConverter/issues/1035)
 
 ### C# -> VB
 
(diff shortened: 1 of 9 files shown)


GrahamTheCoder (approved):
Looks good. The failing tests probably just assert the old behaviour, which we now know is unnecessary, so you can just run them with recharacterize set to true.

Yozer:
Yep, that's what I did.
I also compiled the new tests results to make sure that the not-renamed enum doesn't break them.

GrahamTheCoder:
Hmm yes, good plan to check since I don't think the MultiFileCharacterization ones flag anywhere if the compilation of the result fails (whereas the normal ones output compile errors at the bottom of the output)

PR #1060, opened 2023-12-09
Latest vs/multi file tests

GrahamTheCoder:
With the latest version of VS I was having difficulty running the multi-file tests.
They're a bit delicate since they end up with the test runner, codeconv.netframework, installed VS and installed dotnet sdk all vying to set versions of things on a solution which has a wide variety of project types

One of the easier bits to workaround looked suspiciously like this bug marked as fixed: https://github.com/dotnet/roslyn/issues/64323


GrahamTheCoder:
Interesting there was a patch for that which should be in 4.8, but the issue was never closed: https://github.com/dotnet/roslyn/issues/63921

Issue #1042, opened 2023-11-16
Move Functions to net80 isolated model


christophwille:
https://techcommunity.microsoft.com/t5/apps-on-azure-blog/azure-functions-ignite-2023-update/ba-p/3978641

In our case a very minor change, but still we should track it. (maybe easiest to do File/New Project and then copy over)

PR #1079, opened 2024-01-30
Upgrade Azure Function to .NET 8

christophwille:
Implement #1042


GrahamTheCoder:
Nice one, I was just thinking about this having upgraded some other projects I work on yesterday

christophwille:
@GrahamTheCoder any ideas why a unit test would fail although I didn't touch them at all? Locally in VS the unit tests run just fine.

GrahamTheCoder:
I think there is a bit of non-determinism in the roslyn simplifier (or possibly it's hitting a timeout). It on rare occasions just decides to leave "global" on a few bits on one of the outputs. I've never spotted a pattern to when it happens and running again always fixes.

Issue #172, opened 2018-08-27
Awaiting user request: Convert ASP.NET (+core) cshtml/vbhtml files.


GrahamTheCoder:
May be necessary to use things like 
* https://docs.microsoft.com/en-us/dotnet/api/system.web.compilation.buildprovider?redirectedfrom=MSDN&view=netframework-4.7.2, 
* https://github.com/dotnet/roslyn/issues/11324#issuecomment-263020905 
to get the syntax tree, or may already be in the syntax trees and just need special treatment.

Out of scope: ASP Net Classic [will not be supported](https://github.com/icsharpcode/CodeConverter/issues/160)

GrahamTheCoder:
This solution could probably also cover https://github.com/icsharpcode/CodeConverter/issues/219

camainc:
Has there been movement on this? We have a huge web application with hundreds of vbhtml files and could use a helping hand.

GrahamTheCoder:
Just to check I'm looking at the right thing, am I right in thinking it's .net framework 4.8 that's the latest version that supports vbhtml? Is that what you're using?

I've started adding some background information to the issue description. The good news is that it's possible to convert the code, the less good news is that we nee

GrahamTheCoder:
I've got a spike that gets me access to what's needed I think, but it's reflection heavy: https://github.com/icsharpcode/CodeConverter/pull/1110

PR #1110, opened 2024-06-15
Issue/172

GrahamTheCoder:
Spike relates to #172
Best solution available is probably to match the vbhtml structure against the generated vb document (which follows a simple pattern), then manually call to convert each part of the syntax tree separately so it's easy to construct the final document without any manual parsing - after which the normal converter functions can be used.

I haven't found the correct settings to get the vb document yet. It may even need an older version of the library that targeted dot net framework

### DIFF
--- a/Tests/CSharp/SpikeTests.cs
+++ b/Tests/CSharp/SpikeTests.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using System.Collections.Immutable;
+using System.IO;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis;
+using Xunit;
+
+
+namespace ICSharpCode.CodeConverter.Tests.CSharp;
+
+public class SpikeTests
+{
+
+    [Fact]
+    public async Task ConvertRazorFileAsync()
+    {
+        var codeDocument = GetCodeDocument("""
+        @Code
+            ViewData("Title") = "About" & CStr(1)
+        End Code
+
+        <main aria-labelledby="title">
+            <h2 id="title">@ViewData("Title").</h2><h3 id="title">@ViewData("Title2").</h3>
+            <h3>@ViewData("Message")</h3>
+
+            <p>Use this area to provide additional information.</p>
+        </main>@Code
+            ViewData("Title") = "About" & CStr(1)
+        End Code
+
+        <main aria-labelledby="title">
+            <h2 id="title">@ViewData("Title").</h2><h3 id="title">@ViewData("Title2").</h3>
+            <h3>@ViewData("Message")</h3>
+
+            <p>Use this area to provide additional information.</p>
+        </main>
+
+        """, "anything.vbhtml");
+        
+        var syntaxTree = codeDocument.GetSyntaxTree();
+        var getRoot = syntaxTree.GetType().GetProperty("Root", BindingFlags.Instance | BindingFlags.NonPublic).GetMethod;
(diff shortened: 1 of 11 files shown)


GrahamTheCoder:
It's likely a fair bit of work to get what could turn out to break next version since it depends on internals of another library.
I don't believe there's a public api that allows this. If someone wants to continue this route, I will accept the PR and advise where I can, but I'm not planning to take this any further myself.

Vsix/CodeConversion.cs
```diff
@@ -308,7 +308,8 @@ private async IAsyncEnumerable<ConversionResult> ConvertProjectUnhandledAsync<TL
         var projectsByPath =
             _visualStudioWorkspace.CurrentSolution.Projects.ToLookup(p => p.FilePath, p => p);
 #pragma warning disable VSTHRD010 // Invoke single-threaded types on Main thread - ToList ensures this happens within the same thread just switched to above
-        var projects = selectedProjects.Select(p => projectsByPath[p.FullName].First()).ToList();
+        var projects = selectedProjects.SelectMany(p => projectsByPath[p.FullName]).ToList();
```
GrahamTheCoder, line 311: This was the initial spike that allowed getting the project that's created for the open editor. I didn't pursue this since even with the faff of needing the windows open, it didn't seem to easily have access to the mapped spans to insert the code back into the original vbhtml file from the generated file

Tests/CSharp/SpikeTests.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        """, "anything.vbhtml");
+        
+        var syntaxTree = codeDocument.GetSyntaxTree();
+        var getRoot = syntaxTree.GetType().GetProperty("Root", BindingFlags.Instance | BindingFlags.NonPublic).GetMethod;
+        var root = getRoot.Invoke(syntaxTree, BindingFlags.Instance | BindingFlags.NonPublic, null, null, null);
+        var getDocument = root.GetType().GetProperty("Document", BindingFlags.Instance | BindingFlags.Public).GetMethod;
+        var document = getDocument.Invoke(root, BindingFlags.Instance | BindingFlags.NonPublic, null, null, null);
+        var getChildren = document.GetType().GetProperty("Children", BindingFlags.Instance | BindingFlags.Public).GetMethod;
+        var children = getChildren.Invoke(document, BindingFlags.Instance | BindingFlags.NonPublic, null, null, null);
```
GrahamTheCoder, line 51: I'm pretty sure everything needed is in here, but it will take a lot of brittle reflection calls to internal methods to put it together.

Ultimately, need to get a vb syntaxtree for the whole file, and be able to map each code block back to the vbhtml so we can insert the relevant cs there at the end of conversion.

Tests/CSharp/SpikeTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        var sourceDocument = TestRazorSourceDocument.Create(source, normalizeNewLines: true);
+        var importDocument = TestRazorSourceDocument.Create("@addTagHelper *, TestAssembly", filePath: filePath, relativePath: filePath);
+
+        var codeDocument = engine.ProcessDesignTime(sourceDocument, FileKinds.Legacy, importSources: ImmutableArray.Create(importDocument), new[] { taghelper });
+        return codeDocument;
+        //return RazorWrapperFactory.WrapCodeDocument(codeDocument);
+    }
+    protected RazorProjectEngine CreateProjectEngine(Action<RazorProjectEngineBuilder> configure)
+    {
+        return RazorProjectEngine.Create(RazorConfiguration.Default, RazorProjectFileSystem.Create("c:\\tmp"), configure);
```
GrahamTheCoder, line 83: May need an older version of the library to parse vbhtml, or just to tweak the settings here. If it's an older version, would need to check we're allowed to redistribute it, but it would solve any brittleness issues since it won't change.

Issue #1064, opened 2023-12-13
VB -> C#: improve int const adjustments in some cases


TymurGubayev:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when a hexadecimal constant is converted in a way I can't search for it anymore (in some cases).

VB.NET input:
```vb
Dim i As Integer = &H80040E19
If (i = &H80040E19) Then
End If
```

Current output:
```cs
int i = int.MinValue + 0x00040E19;
if (i == int.MinValue + 0x00040E19)
{
}
```

**Describe the solution you'd like**
I think a much nicer output would be using `unchecked` instead of changing the constant literal, something like this:

```cs
int i = unchecked((int)0x80040E19);
if (i == unchecked((int)0x80040E19))
{
}
```

**Describe alternatives you've considered**
Write a Roslyn replacement rule to restore the constants in a pass after CodeConverter is done.

**Additional context**
None.

TymurGubayev:
resolved via #1115

PR #1115, opened 2024-07-18
generate `unchecked((int)0x91234567)` instead of `int.MinValue + 0x11234567`

TymurGubayev:
### Problem
#1064
For `&H80040E19` we get `int.MinValue + 0x00040E19`. With this PR, we'd get `unchecked((int)0x80040E19)` instead, thus preserving the hex literal value.

### Solution
Corresponding code in LiteralConversion.cs was adjusted.

Additionally, I had to change `CsWouldBeSimplifiedIncorrectly` to return true for `CastExpressionSyntax` with `Parent` that's a `CheckedExpressionSyntax`, because otherwise it would've caused CS0266: "Cannot implicitly convert type 'uint' to 'int'".
Note this rule is a bit too aggressive:
```cs
int i = unchecked(0x80040E19); //we prevent error CS0266 here
uint u = unchecked(0x80040E19); //but we also prevent this from happening
```
(and many more cases with different types)

I'm reasonable sure it doesn't matter though: this PR introduces the very first call to `SyntaxFactory.CheckedExpression`.

### DIFF
--- a/CodeConverter/CSharp/LiteralConversions.cs
+++ b/CodeConverter/CSharp/LiteralConversions.cs
@@ -152,18 +152,16 @@ internal static class LiteralConversions
             int parsedHexValue = int.Parse(hexValue, NumberStyles.HexNumber, CultureInfo.InvariantCulture);
 
             // This is a very special case where for 8 digit hex strings, C# interprets them as unsigned ints, but VB interprets them as ints
-            // This can lead to a compile error if assigned to an int in VB. So in a case like 0x91234567, we generate `int.MinValue + 0x11234567`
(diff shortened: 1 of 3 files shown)


TymurGubayev:
the unit test that failed is unrelated, with following difference:
```cs
[System.Runtime.CompilerServices.CompilerGenerated()]
[global::System.Runtime.CompilerServices.CompilerGenerated()]
```

I had it locally once, and I think it's because of the modal dialog `Debug.Fail` wants to show (`CachedReflectedDelegates.cs`, line 60).

TymurGubayev:
watched the unit test error some more, it just happens sporadically from time to time.

GrahamTheCoder:
Interesting on the debug fall, I've never caught the root cause of that transient issue

Issue #1067, opened 2023-12-19
VB -> C#: incorrect output code if ref return from a C# lib is involved


TymurGubayev:
### VB.Net input code
We need a C#-project with a `ref return` first, f.e.
```cs
public class MySpecialList<T>
{
    private T dummy;
    public ref T this[int i]
    {
        get
        {
            return ref dummy;
        }
    }
}
```

Then in vb.net (array is for comparison)
```vbnet
Sub UseArr()
    Dim arr() As Object
    Modify(arr(0))
End Sub

Sub UseRefReturn()
    Dim lst As CSProj.MySpecialList(Of Object)
    Modify(lst(0))
End Sub

Sub Modify(ByRef o As Object)
End Sub
```

### Erroneous output
```C#
public void UseArr()
{
    var arr = default(object[]);
    Modify(ref arr[0]);
}

public void UseRefReturn()
{
    var lst = default(CSProj.MySpecialList<object>);
    var tmp = lst;
    var argo = tmp[0];
    Modify(ref argo); //note the modified value never makes it back into the list
}

public void Modify(ref object o)
{
}
```

### Expected output
```C#
public void UseArr()
{
    var arr = default(object[]);
    Modify(ref arr[0]);
}

public void UseRefReturn()
{
    var lst = default(CSProj.MySpecialList<object>);
    Modify(ref lst[0]);
}

public void Modify(ref object o)
{
}
```

### Details
* Product in use: VS extension
* Version in use: 9.2.4 (dd3edd8862e6a743eca532c861145768c6601929)
* Did you see it working in a previous version, which? No
* Any other relevant information to the issue, or your interest in contributing a fix.

This is most probably related to https://github.com/icsharpcode/CodeConverter/issues/1052#issuecomment-1856145388

TymurGubayev:
fixed in #1116

Issue #1052, opened 2023-12-05
VB -> C#: `With Me` generates incorrect code (CS1605)


TymurGubayev:
### VB.Net input code
```vbnet
Public Class TestWithMe
    Private _x As Integer
    Sub S()
        With Me
            ._x = 1
        End With
    End Sub
End Class
```

### Erroneous output
```C#
    public class TestWithMe
    {
        private int _x;
        public void S()
        {
            {
                ref var withBlock = ref this; //CS1605: Cannot use 'this' as a ref or out value because it is read-only
                withBlock._x = 1;
            }
        }
    }
```

### Expected output
I think we can omit `ref` for reference types. Unfortunately, the `With Me`-syntax is valid for structures as well... So, maybe in that case don't create the `withBlock`-Variable and use `this` instead?

```C#
    public class TestWithMe
    {
        private int _x;
        public void S()
        {
            {
                var withBlock = this;
                withBlock._x = 1;
            }
        }
    }
```

### Details
* Product in use: VS extension
* Version in use: ae25cdc920df68755fb233934028091c97b42853
* Did you see it working in a previous version, which? No
* Any other relevant information to the issue, or your interest in contributing a fix.

GrahamTheCoder:
As the error message implies, the best solution to this would be one that detected whether the identifier referred to a readonly symbol in general (e.g. a readonly field). There's actually a special separate compiler error for trying to use a loop iteration variable in this way for example too:

```csharp
using System.Collections.Generic;

public struct TestWithMe
{
    private int _x;
    public static void Main()
    {
        new TestWithMe().M();
    }

    public void M()
    {

        var x = new List<TestWithMe>();
        foreach(var y in x){
            ref var withBlock =  ref y; // CS1657: Cannot use 'y' as a ref or out value because it is a 'foreach iteration variable'
            withBlock._x = 1;
            System.Console.Write(_x);
        }
    }
}
```

TymurGubayev:
a similar problem arises if the `withBlock` is used inside a lambda created to resolve constants passed to ByRef parameters.

TymurGubayev:
I've just noticed, output code in https://github.com/icsharpcode/CodeConverter/issues/1049 is actually missing the `ref` and so is semantically incorrect (the struct that's changed is the local copy instead of the instance in the special array thing):

Current output:
```cs
{
    var withBlock = lst[i];
    withBlock.SomeField = 5;
    s = withBlock.Text;
}
```
Should've been:
[...]

PR #1116, opened 2024-07-19
Handle C# `ref return` when converting VB->C#

TymurGubayev:
### Problem
* #1067: VB -> C#: incorrect output code if `ref return` from a C# lib is involved
* Related problem described in this [comment](https://github.com/icsharpcode/CodeConverter/issues/1052#issuecomment-1856145388): `var withBlock = lst[i];` should be `ref var withBlock = ref lst[i];` if `ref return` from a C# lib is involved

### Solution
* The changes in code are actually minor: I added checks for `prop.ReturnsByRef || prop.ReturnsByRefReadonly` in a few places.
   I had to ignore a project in `ConvertWholeSolutionAsync` (C#->VB) because of the `ref return`.
* Currently, the `MultiFileSolutionAndProjectTests` tests **will fail** because of  https://github.com/dotnet/roslyn/issues/68194
  This is fixed in Roslyn 4.8, but I haven't touched the references.


GrahamTheCoder:
There's now a prerelease that seems to generally work, so I've pushed that.
The test "OutOptionalArgumentAsync" no longer detects the out parameters correctly in the new version (unrelated to your changes). This means people using that version of VS will see this regression, but it's not related to your change, so I'll skip the test and merge this.

GrahamTheCoder:
Just spotted the project level tests are having an issue too with the new version of roslyn

TymurGubayev:
My findings so far:

* upgrading Microsoft.Build[.Tasks.Core] 17.4.0 -> 17.10.4 but leaving Roslyn at 4.1.0 leaves us with the `System.InvalidOperationException : Unexpected value 'PropertyAccess' of type 'Microsoft.CodeAnalysis.VisualBasic.BoundKind'`.

* upgrading Roslyn 4.1.0 -> 4.8.0 causes `Cannot convert SimpleImportsClauseSyntax, System.InvalidOperationException: Service of type 'Microsoft.CodeAnalysis.Classification.IEmbeddedLanguageClassificationService' is required to accomplish the task but is not available from the workspace.` (i.e. the https://github.com/dotnet/roslyn/issues/63921)

* upgrading Roslyn 4.8.0 -> 4.9.2 (and further)  causes `System.InvalidOperationException : Unable to locate the .NET Framework build host at C:\Users\username\AppData\Local\Temp\d087edd0-b2a6-452c-9149-94c9f59c027b\d087edd0-b2a6-452c-9149-94c9f59c027b\assembly\dl3\9fbdf688\000bbd93_706bda01\BuildHost-net472\Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.exe - line 206`

The path it's looking at is `Path.Combine(Path.GetDirectoryName(typeof(BuildHostProcessManager).Assembly.Location), "BuildHost-net472", "Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.exe")` (from Roslyn's BuildHostProcessManager.cs, `GetDotNetFrameworkBuildHostPath()`)
Everything up to `/BuildHost-net472/` exists.

To me this looks like a Roslyn issue, but I have no idea what's going on there.

TymurGubayev:
I think the problem is in the combination of Xunit and some change in Roslyn.
With Roslyn 4.8.0, the `Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.EXE` is loaded from `%temp%\...guids...\assembly\dl3\<id1>\<id2>\Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.exe`, not from a `BuildHost-net472/`-subfolder.

Also, a similar setup using MSTest works fine.

GrahamTheCoder:
OK yeah looks like we're stuck. Let's merge this fix with the tests marked as skipped and a comment linking to this issue or directly to the upstream bugs if/when logged upstream.

TymurGubayev:
Btw, I tried following workaround:

```cs
private void CopyBuildHost()
{
    Assembly[] loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies();
    Assembly msBuildAssembly = Array.Find(loadedAssemblies, assembly => assembly.FullName.Contains("Microsoft.CodeAnalysis.Workspaces.MSBuild"));

    var BuildHostDirName = "BuildHost-net472";
    if (!Directory.Exists($"./{BuildHostDirName}")) return;

    var targetFolder = Path.Combine(Path.GetDirectoryName(msBuildAssembly.Location), BuildHostDirName);
        
    Directory.CreateDirectory(targetFolder);
    foreach (string file in Directory.GetFiles(BuildHostDirName))
    {
        string fileName = Path.GetFileName(file);
        string destinationPath = Path.Combine(targetFolder, fileName);
        File.Copy(file, destinationPath, false);
    }
}
```

which got me `System.IO.IOException : The pipe has been ended.` instead.

GrahamTheCoder:
I was getting `System.IO.IOException : The pipe has been ended.` on the latest version the whole time, so maybe something is a bit different on your local setup. e.g. I run the tests through ReSharper's test runner which may have made a difference

Issue #1048, opened 2023-11-29
VB -> C#: Interop method invocations with OptionalAttribute aren't converted


TymurGubayev:
### VB.Net input code
```vbnet
    Sub S()
        Dim what = Microsoft.Office.Interop.Word.WdGoToItem.wdGoToLine

        Dim objRange As Microsoft.Office.Interop.Word.Range
        objRange = objRange.GoTo(what, , -1)
        objRange = objRange.GoTo(what, Nothing, -1)
    End Sub
```

where the signature of the GoTo method is

```cs
    [MethodImpl(MethodImplOptions.InternalCall, MethodCodeType = MethodCodeType.Runtime)]
    [DispId(173)]
    [return: MarshalAs(UnmanagedType.Interface)]
    Range GoTo([Optional][In][MarshalAs(UnmanagedType.Struct)] ref object What, [Optional][In][MarshalAs(UnmanagedType.Struct)] ref object Which, [Optional][In][MarshalAs(UnmanagedType.Struct)] ref object Count, [Optional][In][MarshalAs(UnmanagedType.Struct)] ref object Name);
```

### Erroneous output
```C#
        public void S()
        {
            var what = Microsoft.Office.Interop.Word.WdGoToItem.wdGoToLine;

            var objRange = default(Microsoft.Office.Interop.Word.Range);
            object argWhat = what;
            object argCount = -1;
            objRange = default
#error Cannot convert InvocationExpressionSyntax - see comment for details
   /* Cannot convert ArgumentListSyntax, System.InvalidOperationException: Operation is not valid due to the current state of the object.
      at Microsoft.CodeAnalysis.VisualBasic.Symbols.ParameterSymbol.get_ExplicitDefaultValue()
      at ICSharpCode.CodeConverter.CSharp.ExpressionNodeVisitor.CreateOptionalRefArg(IParameterSymbol p, RefKind refKind)
      at ICSharpCode.CodeConverter.CSharp.ExpressionNodeVisitor.<GetAdditionalRequiredArgs>d__132.MoveNext()
      at System.Linq.Enumerable.<ConcatIterator>d__59`1.MoveNext()
      at Microsoft.CodeAnalysis.CSharp.SyntaxFactory.SeparatedList[TNode](IEnumerable`1 nodes)
      at ICSharpCode.CodeConverter.CSharp.ExpressionNodeVisitor.<VisitArgumentList>d__55.MoveNext()
[...]

PR #1122, opened 2024-07-22
take parameter attributes into account in `CreateOptionalRefArg`

TymurGubayev:
### Problem
#1048: Interop method invocations with OptionalAttribute aren't converted

### Solution
* *Any comments on the approach taken, its consistency with surrounding code, etc.*
  I decided to leave compiler errors in code, so that `BC30455: Argument not specified for parameter...` becomes `CS7036: There is no argument given that corresponds to the required formal parameter...`.
* *Which part of this PR is most in need of attention/improvement?*
  1. I couldn't quite achieve above goal: VB.NET doesn't respect the `Optional` parameter (because it doesn't need to, I guess), so we have 3 compiler errors in the VB.NET code instead of 1 should we reference a C# assembly. This means the output generated looses some compiler errors -- I'm not quite sure if that's ok. We'd need to setup two different projects to have a proper test, which I'm not too keen on doing because of the problems with #1116.
  2. For parameters without `DefaultParameterValueAttribute`, I use `default` as initializer. I think this is ok...
  3. I'm using `_semanticModel.Compilation.GetTypeByMetadataName("System.Runtime.InteropServices.DefaultParameterValueAttribute");` to get the attribute class (to keep it consistent with other places). If it fails to find one, I'm assuming it's not present -- which isn't necessarily true, I think, if at all possible. Alternative would be something like this:
   ```cs
    //var attributeData = p.GetAttributes().FirstOrDefault(a => SymbolEqualityComparer.IncludeNullability.Equals(a.AttributeClass, defaultParameterValueAttribute));
    var attributeData = p.GetAttributes().FirstOrDefault(a => a.AttributeClass.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat) == "global::System.Runtime.InteropServices.DefaultParameterValueAttribute");
    ```
    This is IMO marginally safer, removes a call to `GetTypeByMetadataName`, but is maybe uglier.

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -35,7 +35,6 @@ internal class ExpressionNodeVisitor : VBasic.VisualBasicSyntaxVisitor<Task<CSha
     private readonly QueryConverter _queryConverter;
     private readonly Lazy<IReadOnlyDictionary<ITypeSymbol, string>> _convertMethodsLookupByReturnType;
     private readonly LambdaConverter _lambdaConverter;
-    private readonly INamedTypeSymbol _vbBooleanTypeSymbol;
     private readonly VisualBasicNullableExpressionsConverter _visualBasicNullableTypesConverter;
     private readonly Dictionary<string, Stack<(SyntaxNode Scope, string TempName)>> _tempNameForAnonymousScope = new();
     private readonly HashSet<string> _generatedNames = new(StringComparer.OrdinalIgnoreCase);
@@ -58,7 +57,6 @@ internal class ExpressionNodeVisitor : VBasic.VisualBasicSyntaxVisitor<Task<CSha
         // If this isn't needed, the assembly with Conversions may not be referenced, so this must be done lazily
         _convertMethodsLookupByReturnType =
             new Lazy<IReadOnlyDictionary<ITypeSymbol, string>>(() => CreateConvertMethodsLookupByReturnType(semanticModel));
-        _vbBooleanTypeSymbol = _semanticModel.Compilation.GetTypeByMetadataName("System.Boolean");
     }
 
     private static IReadOnlyDictionary<ITypeSymbol, string> CreateConvertMethodsLookupByReturnType(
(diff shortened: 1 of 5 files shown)


CodeConverter/CSharp/ExpressionNodeVisitor.cs
```diff
@@ -1818,8 +1818,54 @@ private ArgumentSyntax CreateExtraArgOrNull(IParameterSymbol p, bool requiresCom
...
+            }
+
+            return p.GetAttributes().Any(a => SymbolEqualityComparer.IncludeNullability.Equals(a.AttributeClass, optionalAttribute));
+        }
+    
+        bool TryGetDefaultParameterValueAttributeValue(IParameterSymbol p, out object defaultValue)
+        {
+            defaultValue = null;
+
+            var defaultParameterValueAttribute = _semanticModel.Compilation.GetTypeByMetadataName("System.Runtime.InteropServices.DefaultParameterValueAttribute");
```
GrahamTheCoder, line 1852: The implementation only caches the last 50 requested types, and they can be a bit expensive to find, so I tend to try to stash them in a field if they may be hit a lot like these.

GrahamTheCoder (approved):
Yep looks good. At some point it'd be good to PR an udpate to the changelog, but obviously while there are several PRs flying it will just cause unncessary conflicts so it's fine for now

CodeConverter/CSharp/ExpressionNodeVisitor.cs
TymurGubayev: and what about using `.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat)`?

GrahamTheCoder: Haven't even noticed that one sticking out as slow.

This tweak was borderline premature optimisation so don't worry too much about it. The slowest things off the top of my head are formatting a file, the GetOperation API and any analysis e.g. control flow analysis.
Any micro optimisations like which order you check some already-calculated booleans are totally irrelevant compared to those big hitters so don't worry too much so long as you aren't introducing n^2 algorithms

I haven't done any profiling for a while, I should probably profile the tests some time to see if there are any quick wins, but profiling on a real code base would be even better.

TymurGubayev: that (performance), but also following possible issues:

1. The code does checks like this:
```cs
var optionalAttribute = CommonConversions.KnownTypes.OptionalAttribute;
if (optionalAttribute == null) {
    return false;
}
```

But if `optionalAttribute` is `null`, something went terribly wrong: probably the compilation doesn't have the class for some reason... Maybe we should do something drastic in this case?

2. Things like `SymbolEqualityComparer.IncludeNullability.Equals(a.AttributeClass, defaultParameterValueAttribute)` work only if the class symbols are from the same compilation, IIRC... I hope that's always the case, but I haven't look too much into the details.

Issue #1054, opened 2023-12-05
VB -> C#: Statics with same name in getter and setter cause a CONVERSION ERROR


TymurGubayev:
### VB.Net input code
```vbnet
    Public Property Prop As String
        Get
            Static b As Boolean
            b = False
        End Get

        Set(ByVal s As String)
            Static b As Boolean
            b = False
        End Set
    End Property
```

### Erroneous output
```C#
#error Cannot convert ClassBlockSyntax - see comment for details
    /* Cannot convert ClassBlockSyntax, CONVERSION ERROR: An item with the same key has already been added. in 'Public Class TestStaticsInS...' at character 0
       at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
       at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
       at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
       at ICSharpCode.CodeConverter.CSharp.PerScopeState.<CreateVbStaticFieldsAsync>d__20.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
       at ICSharpCode.CodeConverter.CSharp.DeclarationNodeVisitor.<>c__DisplayClass40_0.<<ConvertMembersAsync>b__1>d.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
       at ICSharpCode.CodeConverter.Common.AsyncEnumerableTaskExtensions.<SelectAsync>d__3`2.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
       at ICSharpCode.CodeConverter.Common.AsyncEnumerableTaskExtensions.<SelectManyAsync>d__0`2.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
       at ICSharpCode.CodeConverter.CSharp.DeclarationNodeVisitor.<ConvertMembersAsync>d__40.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
[...]

PR #1124, opened 2024-07-23
Differentiate between statics in setters and getters

TymurGubayev:
### Problem
#1054: Statics with same name in getter and setter cause a CONVERSION ERROR

### Solution
  The solution is basically to store accessor kind in every `IdentifierNameSyntax` of a static, and to use that value later when doing the renaming.
  Currently, it creates `_Prop_b` and `_Prop_b1`, which is fine, but it would be nicer to get `_Prop_Get_b` and `_Prop_Set_b` instead. This is easy to achieve if *every* static from a property would get that infix. It's a bit more cumbersome if only those with duplicate name would get one.

### DIFF
--- a/CodeConverter/CSharp/HoistedFieldFromVbStaticVariable.cs
+++ b/CodeConverter/CSharp/HoistedFieldFromVbStaticVariable.cs
@@ -22,5 +22,5 @@ internal class HoistedFieldFromVbStaticVariable : IHoistedNode
     }
 
     public string FieldName => OriginalMethodName != null ? $"_{OriginalMethodName}_{OriginalVariableName}" : $"_{OriginalVariableName}";
-    public string PrefixedOriginalVariableName => PerScopeState.GetPrefixedName(OriginalParentAccessorKind.ToString(), OriginalVariableName);
+    public string PrefixedOriginalVariableName => PerScopeState.GetPrefixedName(OriginalParentAccessorKind == MethodKind.Ordinary ? "" : OriginalParentAccessorKind.ToString(), OriginalVariableName);
 }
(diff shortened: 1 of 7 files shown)


CodeConverter/CSharp/HoistedFieldFromVbStaticVariable.cs
```diff
@@ -6,18 +6,21 @@ internal class HoistedFieldFromVbStaticVariable : IHoistedNode
...
     public ExpressionSyntax Initializer { get; }
     public TypeSyntax Type { get; }
     public bool IsStatic { get; }
 
-    public HoistedFieldFromVbStaticVariable(string originalMethodName, string originalVariableName, ExpressionSyntax initializer, TypeSyntax type, bool isStatic)
+    public HoistedFieldFromVbStaticVariable(string originalMethodName, string originalVariableName, string originalParentAccessorKind, ExpressionSyntax initializer, TypeSyntax type, bool isStatic)
     {
         OriginalMethodName = originalMethodName;
         OriginalVariableName = originalVariableName;
+        OriginalParentAccessorKind = originalParentAccessorKind;
```
GrahamTheCoder, line 18: As a general rule, try to keep strong typing where it exists. Much easier for the receiver to know they have that enum and can do whatever with it rather than just an opaque string.

In the general case, I mean keep an enum or a small wrapper, I don't mean keep the whole syntax node around.

If keeping it as a string here feels right, then the name is probably wrong, and it should just be something like UniquePrefixFromContext here. Either way is OK, it's just this walks the middle path where both the caller and consumer are making half the decision about how to use this.

TymurGubayev: I chose string because `SyntaxAnnotation.Data` is a string. And I thought I'd avoid parsing the data back to an enum, but it doesn't happen anyway with the way the code currently is, so I replaced it with the `MethodKind` enum now.

TymurGubayev: The choice is between `VBasic.SyntaxKind`, `MethodKind` and some new enum just for that field... If you prefer another option, I'll change it.

GrahamTheCoder: Looks good, thanks!

PR #1125, opened 2024-07-24
Don't write back into a method call when it's ByRef argument

TymurGubayev:
### Problem
ByRef argument w/o parentheses is treated like a writable variable, even if it's a method:

```vb
Sub S(ByRef i as Object)
    M(GetI)
End Sub
Function GetI() As Integer : End Function
```
becomes
```cs
public bool M(ref object i)
{
    object argi1 = GetI();
    M(ref argi1);
    GetI() = argi1; //compiler error
}

public int GetI()
{
    return default;
}
```

### Solution
  The adjustment itself is very minor: return `RefConversion.PreAssigment` from `GetRefConversion` when the symbol is an `IMethodSymbol`.
  I've excluded `ReturnsByRef == false && ReturnsByRefReadonly == false`, same as with properties. This generates code like this atm:
   ```cs
   object argi2 = byrefThis.RefProp;
   M(ref argi2);
   byrefThis.RefProp = Conversions.ToInteger(argi2);

   object argi3 = byrefThis.GetRefVal();
   M(ref argi3);
   byrefThis.GetRefVal() = Conversions.ToInteger(argi3);
   ```
   I haven't added any tests for this though, because of the whole problematic with solution tests involving C# ref returns from VB.NET.

   Also, this code isn't necessarily correct if the property or method involved has side effects. It should use a temporary ref variable, something like this:
   ```cs
   ref int tmp = ref byrefThis.GetRefVal();
   object argi3 = tmp;
   M(ref argi3);
   tmp = (int)(argi3);
   ```
   Unless there is an easy way to do that, I'd say it's out of scope of this PR.

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -1884,14 +1884,14 @@ internal class ExpressionNodeVisitor : VBasic.VisualBasicSyntaxVisitor<Task<CSha
         RefConversion GetRefConversion(VBSyntax.ExpressionSyntax expression)
         {
             var symbolInfo = GetSymbolInfoInDocument<ISymbol>(expression);
-            if (symbolInfo is IPropertySymbol propertySymbol
-            // a property in VB.NET code can be ReturnsByRef if it's defined in a C# assembly the VB.NET code references
-            && !propertySymbol.ReturnsByRef && !propertySymbol.ReturnsByRefReadonly) {
+            if (symbolInfo is IPropertySymbol { ReturnsByRef: false, ReturnsByRefReadonly: false } propertySymbol) {
+                // a property in VB.NET code can be ReturnsByRef if it's defined in a C# assembly the VB.NET code references
                 return propertySymbol.IsReadOnly ? RefConversion.PreAssigment : RefConversion.PreAndPostAssignment;
             }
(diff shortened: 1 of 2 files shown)


GrahamTheCoder:
I think there is some existing code that uses ref locals. I'll try to find it when I have a minute.

GrahamTheCoder:
This is the code I was thinking of in case it helps: https://github.com/icsharpcode/CodeConverter/blob/9e929e1eae1b756564fc8a5362c78d7a9470d24f/CodeConverter/CSharp/MethodBodyExecutableStatementVisitor.cs#L878-L881
Test example: https://github.com/icsharpcode/CodeConverter/blob/9e929e1eae1b756564fc8a5362c78d7a9470d24f/Tests/CSharp/StatementTests/MethodStatementTests.cs#L452

Just a random thought on the issues of testing: I wonder whether we can get around the error by referencing a compiled binary rather than source code which contains a ref return. Either a ref return in the base class library or one we compile manually.

I'll merge this as-is since it seems like an improvement over what's there now.

TymurGubayev:
> Just a random thought on the issues of testing: I wonder whether we can get around the error by referencing a compiled binary rather than source code which contains a ref return. Either a ref return in the base class library or one we compile manually.

I checked what's present in the CharacterizationTestSolution, and it's `System.Span`-related things from netstandard.dll.
I'd prefer custom binary though, because it'd give us more control, and we could maybe use it for Interop-related tests as well.

PR #1126, opened 2024-07-24
Pass ByRef parameters of parent into hoisted function by ref

TymurGubayev:
### Problem
Input:
```vb
Sub S(ByRef x As Integer)
    Dim i As Integer = 0
    If F1(x, i) Then
    End If
End Sub

Function F1(x As Integer, ByRef o As Object) As Boolean : End Function
```

Erroneous output:
```cs
public void S(ref int x)
{
    int i = 0;
    //CS1628: Cannot use ref, out, or in parameter 'x' inside an anonymous method, lambda expression, query expression, or local function
    bool localF1() { object argo = i; var ret = F1(x, ref argo); i = Conversions.ToInteger(argo); return ret; }

    if (localF1())
    {
    }
}

public bool F1(int x, ref object o)
{
    return default;
}
```

### Solution
  Analyze arguments in `HoistAndCallLocalFunctionAsync`, if any are ByRef parameters, create an ArgumentList an a ParameterList to use in the local function we are about to create.
  Currently, I'm passing `null` for the `ParameterList` from `CommonConversions.cs/ConvertEqualsValueClauseSyntaxAsync()` - so there is no change in behavior. It might be worth analyzing what happens there and maybe make a similar adjustment (I haven't spent much time on that).


CodeConverter/CSharp/CommonConversions.cs
```diff
@@ -145,7 +145,7 @@ public record VariablesDeclaration(CSSyntax.VariableDeclarationSyntax Decl, ITyp
                     equalsValueClauseSyntax = null;
                 } else {
                     var returnBlock = SyntaxFactory.Block(SyntaxFactory.ReturnStatement(adjustedInitializerExpr));
-                    _typeContext.PerScopeState.Hoist(new HoistedParameterlessFunction(GetInitialValueFunctionName(vbName), csTypeSyntax, returnBlock));
+                    _typeContext.PerScopeState.Hoist(new HoistedFunction(GetInitialValueFunctionName(vbName), csTypeSyntax, returnBlock, null));
```
GrahamTheCoder, line 148: Agreed this null is slightly unpleasant. More so the field that ends up being null - which I generally wouldn't expect without being indicated by name/annotation (I don't mind having nullable enable at the top of some files, just a wholesale move to it is too much work). I'm surprised you didn't just create the empty parameterlist here.

Another option would have been to make it something like an IReadOnlyList of the parameters that gets passed and stored away (passing through the minimum information needed, and left it to the output methods (e.g. AsLocalFunction) to decide how to format it.

Feel free to tweak it further if either of those sound better to you, but it's a pretty minor issue

TymurGubayev: > I'm surprised you didn't just create the empty parameterlist here.
I kind of forgot that's an option :)

Issue #1145, opened 2024-10-31
Tests are failing if the current culture is not en-US 


LucTremblay:
Hello,

If the current computer culture is not en-US  a lot of asserts will fail, because the language of the error might not be english. 

Here is a example of a failing test:

Expected a message in english
```
Expected: using System;
          
          internal partial class TestClass
          {
              private void TestMethod()
              {
                  if (MyEvent is not null)
                      MyEvent(this, EventArgs.Empty);
              }
          }
          1 source compilation errors:
          BC30451: 'MyEvent' is not declared. It may be inaccessible due to its protection level.
          1 target compilation errors:
          CS0103: The name 'MyEvent' does not exist in the current context

```

The message is in french
```
Actual:   using System;
          
          internal partial class TestClass
          {
              private void TestMethod()
              {
                  if (MyEvent is not null)
                      MyEvent(this, EventArgs.Empty);
              }
          }
          1 source compilation errors:
          BC30451: 'MyEvent' n'est pas déclaré. Il peut être inaccessible en raison de son niveau de protection.
          1 target compilation errors:
          CS0103: Le nom 'MyEvent' n'existe pas dans le contexte actuel
```

A simple fix could be to set the DefaultThreadCurrentUICulture, in the base class of the unit tests seems to fix the issue:
https://github.com/icsharpcode/CodeConverter/compare/master...LucTremblay:CodeConverter:defaultThreadCurrentUICulture

`CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;`
or
`CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-US")`

GrahamTheCoder:
Good point, please PR with invariamt culture if that works, thanks!

PR #1150, opened 2024-11-04
Set the DefaultThreadCurrentUICulture to a specific culture, asserted error language consistency

LucTremblay:
### Problem
#1145 

### Solution
By setting DefaultThreadCurrentUICulture to CultureInfo.InvariantCulture, in the base class of the unit tests, it ensure that the asserted build error text and actual error text will be in the same language

### DIFF
--- a/Tests/TestRunners/ConverterTestBase.cs
+++ b/Tests/TestRunners/ConverterTestBase.cs
@@ -11,6 +11,7 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.VisualBasic;
 using Xunit;
 using Xunit.Sdk;
+using System.Globalization;
 
 namespace ICSharpCode.CodeConverter.Tests.TestRunners;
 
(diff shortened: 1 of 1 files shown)


Tests/TestRunners/ConverterTestBase.cs
```diff
@@ -28,6 +29,7 @@ public class ConverterTestBase
 
     public ConverterTestBase(string rootNamespace = null)
     {
+        CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
```
GrahamTheCoder, line 32: Nice to have: Haven't had need to try this yet, but may be worth trying putting in a [ModuleInitializer ](https://www.cazzulino.com/module-initializers.html) so it definitely covers the whole test suite rather than some tests running with it and some without.

GrahamTheCoder:
Thanks, I'll try to move it to a module initializer now

Issue #1148, opened 2024-11-04
VB -> C#: Func delegate assignments are wrapped in lambdas


gaschd:
In the erroneous output the assignment of FunctionReturningClass to the Func<ITestObj> delegate is wrapped in an additional lambda expression.

### VB.Net input code
```vbnet
Public Class TestClass
    Private Shared FuncClass As Func(Of TestObjClass) = AddressOf FunctionReturningClass
    Private Shared FuncInterface As Func(Of ITestObj) = AddressOf FunctionReturningClass

    Public Shared Function FunctionReturningClass() As TestObjClass
        Return New TestObjClass()
    End Function
End Class

Public Class TestObjClass
    Implements ITestObj
End Class

Public Interface ITestObj
End Interface
```

### Erroneous output
```C#
 public class TestClass
 {
     private static Func<TestObjClass> FuncClass = FunctionReturningClass;
     private static Func<ITestObj> FuncInterface = () => FunctionReturningClass();

     public static TestObjClass FunctionReturningClass()
     {
         return new TestObjClass();
     }
 }

 public class TestObjClass : ITestObj
 {
 }

 public interface ITestObj
 {
 }
```

### Expected output
```C#
 public class TestClass
 {
     private static Func<TestObjClass> FuncClass = FunctionReturningClass;
     private static Func<ITestObj> FuncInterface = FunctionReturningClass;

     public static TestObjClass FunctionReturningClass()
     {
         return new TestObjClass();
     }
 }

 public class TestObjClass : ITestObj
 {
 }

 public interface ITestObj
 {
 }
```

### Details
* Product in use: VS extension
* Version in use: 9.2.6.0

gaschd:
Although the minimal example will compile, adding a parameter to the func will lead to empty lambda params causing a compile time error

PR #1153, opened 2024-11-05
AddressOf does not wrap compatible signatures in lambdas anymore

gaschd:
https://github.com/icsharpcode/CodeConverter/issues/1148

### Problem
AddressOf delegates are wrapped in lambdas when delegate signatures are not equal but still compatible.
This behaviour is implemented correctly for base and derived classes but not for interfaces.

### Solution
Using InheritsFromOrEquals overload with the option to include interfaces for delegates that widen the return type to an interface that is implemented by the type or narrow the input parameters types to classes implementing the interface.


GrahamTheCoder:
Thanks for the PRs, will try to look on the weekend

Issue #1154, opened 2024-11-06
VB -> C#: Nested classes can have the wrong output types


gaschd:
Attributes of the nuget package https://github.com/manuc66/JsonSubTypes are not converted correctly. 

The type
NAMESPACE.CLASS.ATTRIBUTECLASS
is converted to
NAMESPACE.ATTRIBUTECLASS

Maybe this is a case sensitivity issue, as the class is JsonSubtypes, while the namespace is JsonSubTypes?

### VB.Net input code
```vbnet
<JsonSubTypes.JsonSubtypes.KnownSubType(GetType(TestClass), "$type")>
Public Class TestClass
End Class
```

### Erroneous output
```C#
[JsonSubTypes.KnownSubTypeAttribute(typeof(TestClass), "$type")]
public class TestClass
{
}
```

### Expected output
```C#
[JsonSubTypes.JsonSubtypes.KnownSubType(typeof(TestClass), "$type")]
public class TestClass
{
}
```

The same issue occurs using imports:

### VB.Net input code
```vbnet
Imports JsonSubTypes.JsonSubtypes

<KnownSubType(GetType(TestClass), "$type")>
Public Class TestClass
End Class
```

### Erroneous output
```C#
[JsonSubTypes.KnownSubTypeAttribute(typeof(TestClass), "$type")]
public class TestClass
{
}
```

### Expected output
```C#
using static JsonSubTypes.JsonSubtypes;

[KnownSubType(typeof(TestClass), "$type")]
public class TestClass
{
}
```

The C# -> VB conversion works.

### Details
* Product in use: VS extension
* Version in use: 9.2.6.0
* Can contribute

PR #1155, opened 2024-11-06
Issue 1154 wrong nested types conversion

gaschd:
### Problem
https://github.com/icsharpcode/CodeConverter/issues/1154

Namespaces and types can have the same name with a different casing:
```vbnet
Namespace CaseSensitive1
    Public Class Casesensitive1
        Public Class TestDummyAttribute
            Inherits Attribute
        End Class
    End Class
End Namespace

Namespace Issue1154
    <CaseSensitive1.Casesensitive1.TestDummy>
    Public Class UpperLowerCase
    End Class
End Namespace
```
The vbnode-to-csharpnode replacement searches node names with "StringComparison.OrdinalIgnoreCase".
Due to this way of comparison, the type name will be replaced with the namespace name.

### Solution
Performing the node name search with "StringComparison.Ordinal" first. If no match is found retry with "StringComparison.OrdinalIgnoreCase". IgnoreCase is needed as 100's of tests fail when just doing Ordinal comparison.

It is not optimal that the search is done twice now for a lot of declarations, just to handle this edge case, but checking for duplicates with different casing is probably even slower, as the collections are usually very short.

### DIFF
--- a/CodeConverter/CSharp/CommonConversions.cs
+++ b/CodeConverter/CSharp/CommonConversions.cs
@@ -229,7 +229,9 @@ internal class CommonConversions
 
         return syntax.ReplaceNodes(syntax.DescendantNodes().OfType<CSSyntax.IdentifierNameSyntax>(), (oldNode, _) =>
         {
-            var originalName = originalNames.FirstOrDefault(on => string.Equals(on, oldNode.ToString(), StringComparison.OrdinalIgnoreCase));
+            string oldNodeStr = oldNode.ToString();
+            var originalName = originalNames.FirstOrDefault(on => string.Equals(on, oldNodeStr, StringComparison.Ordinal)) ??
+                               originalNames.FirstOrDefault(on => string.Equals(on, oldNodeStr, StringComparison.OrdinalIgnoreCase));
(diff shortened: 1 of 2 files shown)


CodeConverter/CSharp/CommonConversions.cs
```diff
@@ -229,7 +229,9 @@ private static TypeSyntax WithDeclarationNameCasing(TypeSyntax syntax, ITypeSymb
 
         return syntax.ReplaceNodes(syntax.DescendantNodes().OfType<CSSyntax.IdentifierNameSyntax>(), (oldNode, _) =>
         {
-            var originalName = originalNames.FirstOrDefault(on => string.Equals(on, oldNode.ToString(), StringComparison.OrdinalIgnoreCase));
+            string oldNodeStr = oldNode.ToString();
```
GrahamTheCoder, line 232: Existing issue - doesn't need to be fixed here:
It's a bit odd that this uses ToString on the nodes rather than using the identifier text directly. It's _probably_ the same in the implementation, but it's weird this relies on it.

GrahamTheCoder:
Thanks!

Issue #1097, opened 2024-04-25
VB -> C#:  private partial methods do not compile


jrmoreno1:
### VB.Net input code
```vb.net
Public Class VisualBasicClass
  Partial Private Sub MyMethod(p As Integer)
  End Sub
End Class
```

### Erroneous output
```C#
public partial class VisualBasicClass
{
    private partial void MyMethod(int p)
    {
    }
}
```

### Expected output
```C#
public partial class VisualBasicClass
{
    partial void MyMethod(int p);
}
```

### Details
* Both vsix and web site
* Version in use: 9.2.5.0
* Never saw it working
* Adding the body/braces indicates this is supposed to the implementation not the declaration.

jrmoreno1:
Changing the function CanHaveMethodBody to prefix the second test with !ms.IsPartialDefinition &&  fixes the immediate issue (or at least the unit test I wrote for that passes), not sure if the function IsPartialMethodDefinition should also include || ms.IsPartialDefinition as the final check -- I don't know enough to write a test for that.

Also, I put the unit test in SpecialConversionTest instead of MethodStatementTests (as the method doesn't have any statements).

Thoughts?

jrmoreno1:
@GrahamTheCoder : a bit concerned about the unnecessary change to IsPartialMethodDefinition in the PR.  On one hand it didn’t break any test, and on the other it wasn’t necessary for this scenario, but it seems to me like it would have been wrong.

PR #1189, opened 2025-07-18
VB -> C#: private partial methods do not compile

jrmoreno1:
- Added unit test to SpecialConversionTests (not MethodStatementTests  as there is no statements) -ISymbolExtensions.cs
  - CanHaveMethodBody & IsPartialMethodDefinition
  - Added check for IsPartialDefinition

#1097

### Problem
Partial method declarations are including a body, which makes them partial method IMPLEMENTATIONS without declarations and so the resulting conversion does not compile

### Solution
* Changed the code as little as possible, I think it fits in with the existing code, could be broken up into multiple (shorter) lines, and negative pattern matching could probably be used, but didn't want to change too much.
* Change to IsPartialMethodDefinition is unnecessary for this, not sure it should be included or not.
* There is a test for problem, but there's not a test for the other end of it (i.e. that it correctly converts partial implementations, but I didn't have a problem with those).
* [x ] At least one test covering the code changed

### DIFF
--- a/CodeConverter/Util/ISymbolExtensions.cs
+++ b/CodeConverter/Util/ISymbolExtensions.cs
@@ -75,12 +75,12 @@ internal static class ISymbolExtensions
 
     public static bool CanHaveMethodBody(this ISymbol declaredSymbol)
     {
-        return !(declaredSymbol is IMethodSymbol ms) || ms.PartialImplementationPart == null && !ms.IsExtern;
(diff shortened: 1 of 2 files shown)


jrmoreno1:
@GrahamTheCoder : no test results have changed, but I did add a unit test for it.  Not sure it's where you'll want it.

Issue #1170, opened 2025-03-25
VB -> C#: Brackets needed around ternary expression


Nward42:
### VB.Net input code
```vbnet
dim x as integer
x = if(true, 1, 2) + 3
```

### Erroneous output
```C#
int x;
x = true ? 1 : 2 + 3;
```

### Expected output
```C#
int x;
x = (true ? 1 : 2) + 3;
```

### Details
A simplified example of a gotcha hit during a recent VB->C# project conversion using version 9.2.7.0. Simple enough to fix once found. Otherwise, the convertor seems to have worked perfectly. Cheers

GrahamTheCoder:
Thanks for the bug report and feedback on general success!

GrahamTheCoder:
In the online converter, the case you gave seems to work correctly. For ease of repro, do you have an example that fails when pasted into: https://icsharpcode.github.io/CodeConverter/

Nward42:
I simplified the example too much.  The below yields the issue when pasted into the online converter.
```VB
dim i as integer = 0
dim a(If(i = 1, 2, 3)) as string
```

Converts to:
```C#
int i = 0;
var a = new string[i == 1 ? 2 : 3 + 1];
```

Instead of:
```C#
int i = 0;
var a = new string[(i == 1 ? 2 : 3) + 1];
```

So it looks like the problem is in the addition of 1 for C# array extents.

jrmoreno1:
VB arrays can be zero out one based, looks like yours are one based and the converter is trying to keep compatibility as best it can.  This looks to be either the wrong default (should be zero based) or something that is handled in the vb code before doing a conversion.

jrmoreno1:
@Nward42 : I believe my PR will fix this issue.

PR #1191, opened 2025-07-27
VB -> C#: Brackets needed around ternary expression

jrmoreno1:
- Added unit test to AccessExpressionTest.cs
- CommonConversions.cs
  - use SyntaxFactory.ParenthesizedExpression to ParenthesizedExpression the converted expression if it is a ConditionalExpressionSyntax
- Ran all test, no change in existing test, new test passes after initially failing.

Link to issue(s) this covers

#1170

### Problem
C# ternary expression do not inherently have a termination, unlike VB ternary.  When converted without parameterization, IncreaseArrayUpperBoundExpressionAsync adds 1 to the false result.

### Solution
* Is mutating convertedExpression acceptable, or should it be it's own return?
* [X] All existing test continue to pass
* [x] There is now a test for this scenario.


GrahamTheCoder:
Thanks!

Issue #827, opened 2022-02-06
Rare MyClass cases convert incorrectly


GrahamTheCoder:
1. Property in the interface being implemented non-public setter/getter i.e.:
```vb
Interface IFoo     
   Property Prop As Integer 
End Interface
Class Foo 
   Implements IFoo
   Property Prop As Integer Implements IFoo.Prop
        Get
        End Get
        Friend Set
        End Set
    End Property
End Class
```
The converter will generate a property with an internal setter.

2. Not related with explicit interfaces but: Accessing backing field of the autogenerated property where MyClass is involved:
```vb
Class Foo 
    Overridable Property Prop As Integer = 5

    Sub Test()
        _Prop = 10 ' This should convert to MyClassProp = 10 not to Prop = 10
        Dim isCorrect = MyClass.Prop = 10 ' After conversion this will return 5instead of 10 because we wrote to Child.Prop
    End Sub
End Class
Class Child
    Inherits Foo
    Overrides Property Prop As Integer = 20
End Class
```

```csharp
    public class Foo
    {
        public int MyClassProp { get; set; } = 5;

        public virtual int Prop
        {
            get => MyClassProp;

            set => MyClassProp = value;
        }

        public void Test()
        {
            Prop = 10; // This should convert to MyClassProp = 10 not to Prop = 10
            bool isCorrect = MyClassProp == 10; // After conversion this will return 5 instead of 10 because we wrote to Child.Prop
        }
    }

    public class Child : Foo
    {
        public override int Prop { get; set; } = 20;
    }
```

_Originally posted by @Yozer in https://github.com/icsharpcode/CodeConverter/issues/822#issuecomment-1030053015_

PR #1237, opened 2026-03-25
Fix conversion of backing fields of overridable properties

GrahamTheCoder:
This PR addresses issue #827, ensuring that accessing the backing field of an overridable property is correctly translated to accessing `MyClass[PropertyName]`. It also properly identifies these accesses using semantic analysis to trigger the generation of the `MyClass[PropertyName]` delegating auto-property. Added corresponding unit tests in `MemberTests.cs` to guarantee behavior.

---
*PR created automatically by Jules for task [2635451807914525187](https://jules.google.com/task/2635451807914525187) started by @GrahamTheCoder*

### DIFF
--- a/CodeConverter/CSharp/CommonConversions.cs
+++ b/CodeConverter/CSharp/CommonConversions.cs
@@ -294,6 +294,9 @@ internal class CommonConversions
                     text = "value";
                 } else if (normalizedText.StartsWith("_", StringComparison.OrdinalIgnoreCase) && idSymbol is IFieldSymbol propertyFieldSymbol && propertyFieldSymbol.AssociatedSymbol?.IsKind(SymbolKind.Property) == true) {
                     text = propertyFieldSymbol.AssociatedSymbol.Name;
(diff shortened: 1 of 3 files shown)


CodeConverter/CSharp/CommonConversions.cs
```diff
@@ -294,6 +294,9 @@ public SyntaxToken ConvertIdentifier(SyntaxToken id, bool isAttribute = false, S
                     text = "value";
                 } else if (normalizedText.StartsWith("_", StringComparison.OrdinalIgnoreCase) && idSymbol is IFieldSymbol propertyFieldSymbol && propertyFieldSymbol.AssociatedSymbol?.IsKind(SymbolKind.Property) == true) {
                     text = propertyFieldSymbol.AssociatedSymbol.Name;
+                    if (propertyFieldSymbol.AssociatedSymbol.IsVirtual && !propertyFieldSymbol.AssociatedSymbol.IsAbstract) {
+                        text = "MyClass" + text;
+                    }
```
GrahamTheCoder, line 299: The condition here looks a bit suspicous to me. We should add a test for the case of a virtual nonabstract property that isn't a myclass property to guard against this

PR #1238, opened 2026-03-25
Fix Issue 886: VB -> C# wrong conversion of optional struct or decimal ref parameter

GrahamTheCoder:
This PR fixes an issue where converting `Optional ByRef` VB.NET parameters initialized with `Nothing` for custom structures or numeric `Decimal` defaults emitted C# code that fails to compile due to C# compiler limitations with `[DefaultParameterValue]`.

Changes:
1. `ExpressionNodeVisitor.cs`: For `[DefaultParameterValue]`, `decimal` parameter primitive defaults now properly emit an equivalent constant `int` or `double` valid attribute expression, without the decimal `m` suffix which is disallowed. For generic structs (`TypeKind.Struct` and `SpecialType.None`), no `DefaultParameterValue` attribute will be generated as custom structs with non-null defaults fail natively on C#.
2. `ArgumentConverter.cs`: Whenever default parameters expand for variable definitions around callers (`ref` arguments), non-nullable structs initialized to `null` instead will yield their expected `default` initialization.
3. Added the expected unit test `OptionalByRefParameterAsync886` and verified it passes correctly alongside all project suites.

---
*PR created automatically by Jules for task [15338263271799288852](https://jules.google.com/task/15338263271799288852) started by @GrahamTheCoder*

### DIFF
--- a/CodeConverter/CSharp/ExpressionNodeVisitor.cs
+++ b/CodeConverter/CSharp/ExpressionNodeVisitor.cs
@@ -471,7 +471,22 @@ internal partial class ExpressionNodeVisitor : VBasic.VisualBasicSyntaxVisitor<T
                     CS.SyntaxFactory.Attribute(ValidSyntaxFactory.IdentifierName("Optional")),
                 };
                 if (!node.Default.Value.IsKind(VBasic.SyntaxKind.NothingLiteralExpression)) {
-                    optionalAttributes.Add(CS.SyntaxFactory.Attribute(ValidSyntaxFactory.IdentifierName("DefaultParameterValue"), arg));
+                    if (vbSymbol?.Type?.SpecialType == SpecialType.System_Decimal && defaultExpression is CSSyntax.LiteralExpressionSyntax literalExpr) {
+                        var literalValue = literalExpr.Token.Value;
+                        if (literalValue is decimal decimalVal) {
+                            var isInteger = decimalVal == Math.Round(decimalVal);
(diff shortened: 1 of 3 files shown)


CodeConverter/CSharp/ExpressionNodeVisitor.cs
```diff
@@ -471,7 +471,22 @@ public override async Task<CSharpSyntaxNode> VisitParameter(VBSyntax.ParameterSy
                     CS.SyntaxFactory.Attribute(ValidSyntaxFactory.IdentifierName("Optional")),
                 };
                 if (!node.Default.Value.IsKind(VBasic.SyntaxKind.NothingLiteralExpression)) {
-                    optionalAttributes.Add(CS.SyntaxFactory.Attribute(ValidSyntaxFactory.IdentifierName("DefaultParameterValue"), arg));
+                    if (vbSymbol?.Type?.SpecialType == SpecialType.System_Decimal && defaultExpression is CSSyntax.LiteralExpressionSyntax literalExpr) {
+                        var literalValue = literalExpr.Token.Value;
+                        if (literalValue is decimal decimalVal) {
+                            var isInteger = decimalVal == Math.Round(decimalVal);
+                            var newLiteral = isInteger ? CommonConversions.Literal((int)decimalVal) : CommonConversions.Literal((double)decimalVal);
```
GrahamTheCoder, line 478: Seems a bit risky to cast decimal to double if someone really did need the last half of the precision. We should add a condition to check if the number round trips and error otherwise

PR #1254, opened 2026-04-13
Fix conversion of single-statement multiline lambda to block syntax

### LINKED ISSUE #1012: VB -> C#: Conversion of Parallel loop puts multiline LINQ onto one line
### VB.Net input code
```vbnet
Public Class ConversionTest3
    Private Class MyEntity
        Property EntityId As Integer
        Property Name As String
    End Class
    Private Sub BugRepro()

        Dim entities As New List(Of MyEntity)

        Parallel.For(1, 3, Sub(i As Integer)
                               Dim result As String = (From e In entities
                                                       Where e.EntityId = 123
                                                       Select e.Name).Single
                           End Sub)
    End Sub
End Class
```

### Erroneous output
This appears to do something with null checking (which wasn't in the VB code), and it puts all the LINQ code on a single line (due to the Parallel loop).
```C#
using System.Data;
using System.Linq;
using System.Threading.Tasks;

namespace abc
{
    public class ConversionTest3
    {
        private class MyEntity
        {
            public int EntityId { get; set; }
            public string Name { get; set; }
        }
        private void BugRepro()
        {

            var entities = new List<MyEntity>();

            Parallel.For(1, 3, (i) => { string result = (from e in entities where e.EntityId == 123 select e.Name).Single(); });
        }
    }
}
```

### Expected output
```C#
private void BugRepro()
        {

            var entities = new List<MyEntity>();

            Parallel.For(1, 3, (i) => {
                string result = (from e in entities
                                 where e.EntityId == 123
                                 select e.Name).Single();
            });

        }
```

### Details
* Version in use: 9.2.2.0
* When the Parallel loop contains only a single command of code, like my LINQ query above, conversion puts everything on a single line.  If the Parallel loop contains more than one command inside, conversion keeps them on separate lines as expected.
* This is a simplified repro code

GrahamTheCoder:
Fixes #1012 by ensuring that when a VB.NET MultiLineLambdaExpressionSyntax contains a single statement that is not an ExpressionStatementSyntax or ReturnStatementSyntax (such as a local declaration statement with a multiline LINQ query), it is converted to a block-bodied lambda in C# rather than an arrow expression clause.

Also adds a corresponding characterization test.

### DIFF
--- a/CodeConverter/CSharp/LambdaConverter.cs
+++ b/CodeConverter/CSharp/LambdaConverter.cs
@@ -80,8 +80,17 @@ internal class LambdaConverter
             convertedStatements = convertedStatements.Select(l => l.WithTrailingTrivia(SyntaxFactory.ElasticCarriageReturnLineFeed)).ToList();
             block = SyntaxFactory.Block(convertedStatements);
         } else if (singleStatement.TryUnpackSingleExpressionFromStatement(out expressionBody)) {
-            arrow = SyntaxFactory.ArrowExpressionClause(expressionBody);
+            if (vbNode is VBasic.Syntax.MultiLineLambdaExpressionSyntax && singleStatement is not ExpressionStatementSyntax && singleStatement is not ReturnStatementSyntax) {
+                singleStatement = singleStatement.WithTrailingTrivia(SyntaxFactory.ElasticCarriageReturnLineFeed);
+                block = SyntaxFactory.Block(singleStatement);
+                expressionBody = null;
+            } else {
+                arrow = SyntaxFactory.ArrowExpressionClause(expressionBody);
+            }
         } else {
+            if (vbNode is VBasic.Syntax.MultiLineLambdaExpressionSyntax) {
+                singleStatement = singleStatement.WithTrailingTrivia(SyntaxFactory.ElasticCarriageReturnLineFeed);
+            }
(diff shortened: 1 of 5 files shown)


CodeConverter/CSharp/LambdaConverter.cs
```diff
@@ -80,8 +80,17 @@ public async Task<CSharpSyntaxNode> ConvertAsync(VBSyntax.LambdaExpressionSyntax
             convertedStatements = convertedStatements.Select(l => l.WithTrailingTrivia(SyntaxFactory.ElasticCarriageReturnLineFeed)).ToList();
             block = SyntaxFactory.Block(convertedStatements);
         } else if (singleStatement.TryUnpackSingleExpressionFromStatement(out expressionBody)) {
-            arrow = SyntaxFactory.ArrowExpressionClause(expressionBody);
+            if (vbNode is VBasic.Syntax.MultiLineLambdaExpressionSyntax && singleStatement is not ExpressionStatementSyntax && singleStatement is not ReturnStatementSyntax) {
+                singleStatement = singleStatement.WithTrailingTrivia(SyntaxFactory.ElasticCarriageReturnLineFeed);
```
GrahamTheCoder, line 84: Does WithTrailingTrivia(SyntaxFactory.ElasticCarriageReturnLineFeed); make any difference if omitted?

PR #1255, opened 2026-04-13
Claude/fix issue 557 char string default param

### LINKED ISSUE #557: VB -> C#: You can’t assign char to string directly
### Input code
```vbnet
    Friend Const DlM As Char = "^"c

    Friend Function LeftSideOf(ByVal strDlM As String = DlM) As String
      ...
    End Function
```

### Erroneous output
```C#
internal const char DlM = '^';
internal static string LeftSideOf(string strDlM = DlM)
{
  ...
}
```

### Expected output
```C#
internal const char DlM = '^';
internal static string LeftSideOf(string strDlM = null)
{
  strDlM = strDlM ?? DlM?.ToString();
  ...
}
```

### Details
* Product in use: extension
* Version in use: 8.0.4

In c# you can’t assign char to string directly. In case of method param you should assign it **null** and add smth like:   `strDlM = strDlM ?? DlM?.ToString();` in the method body.

GrahamTheCoder:
Fixes #557

### DIFF
--- a/CodeConverter/CSharp/DeclarationNodeVisitor.cs
+++ b/CodeConverter/CSharp/DeclarationNodeVisitor.cs
@@ -4,7 +4,6 @@ using Microsoft.CodeAnalysis.Classification;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Editing;
-using Microsoft.CodeAnalysis.Simplification;
 using static Microsoft.CodeAnalysis.VisualBasic.VisualBasicExtensions;
 using ICSharpCode.CodeConverter.Util.FromRoslyn;
 using ISymbolExtensions = ICSharpCode.CodeConverter.Util.ISymbolExtensions;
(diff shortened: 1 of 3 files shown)


CodeConverter/CSharp/DeclarationNodeVisitor.cs
```diff
@@ -674,11 +675,64 @@ public override async Task<CSharpSyntaxNode> VisitMethodBlock(VBSyntax.MethodBlo
...
+            if (defaultValueNode == null) continue;
+            if (semanticModel.GetTypeInfo(defaultValueNode).Type?.SpecialType != SpecialType.System_Char) continue;
+
+            var csParam = updatedParams[i];
+            var defaultExpr = csParam.Default?.Value;
+            if (defaultExpr is null) continue;
+
+            // Replace the default value with null
+            updatedParams[i] = csParam.WithDefault(
+                CS.SyntaxFactory.EqualsValueClause(ValidSyntaxFactory.NullExpression));
```
GrahamTheCoder: This pattern of mutating previously set values is not recommended. We should set it to be null at the point of creation, then separately here we should add the null coalesce and use comments to mention the other place. This is obviously also not ideal, but a conceptual dependency from a distance is generally better than a concrete one where someone else might intercept it in between in the same style otherwise.

CodeConverter/CSharp/DeclarationNodeVisitor.cs
```diff
@@ -674,11 +676,75 @@ public override async Task<CSharpSyntaxNode> VisitMethodBlock(VBSyntax.MethodBlo
...
+        if (defaultValueNode.IsKind(VBasic.SyntaxKind.CharacterLiteralExpression)) {
+            var constant = semanticModel.GetConstantValue(defaultValueNode);
+            if (constant.HasValue && constant.Value is char c) {
+                return CS.SyntaxFactory.LiteralExpression(CS.SyntaxKind.CharacterLiteralExpression, CS.SyntaxFactory.Literal(c));
+            }
+        }
+        // For named constant references (e.g. DlM or Module.DlM), build a member access expression.
+        // Strip VB's "Global." prefix (VB global namespace qualifier, has no C# identifier equivalent).
+        // Annotate for simplification so Roslyn reduces e.g. TestModule.DlM → DlM within TestModule.
+        var parts = defaultValueNode.ToString().Trim().Split('.');
```
GrahamTheCoder: Reinventing the wheel

CodeConverter/CSharp/DeclarationNodeVisitor.cs
```diff
@@ -674,11 +676,75 @@ public override async Task<CSharpSyntaxNode> VisitMethodBlock(VBSyntax.MethodBlo
...
+        if (parts.Length > 1 && parts[0] == "Global") parts = parts.Skip(1).ToArray();
+        return ValidSyntaxFactory.MemberAccess(parts).WithAdditionalAnnotations(Simplifier.Annotation);
```
GrahamTheCoder: The Simplifier.Annotation should already be added in the simplification pass

